PR #5, opened 2019-09-02
Make JSON iterator return non-const JsonElement

vpodzime:
The container being iterated over should be read-only (to avoid
issues with iterating over a modifying structure). But the child
objects the iterator returns should be modifiabler/read-write,
changing those is not a problem.


cf-bottom:
Thank you for submitting a pull request! Maybe @olehermanse can review this?

PR #8, opened 2019-09-06
Removed files and unit tests which are in core

olehermanse:


olehermanse:
> Would be nice to have a green CFEngine build with these changes before merging.

@vpodzime I generally prefer to wait until it's merged, and use the merge commit in a core PR, but I created this PR for this case specifically: https://github.com/cfengine/core/pull/3826

We should find time to add a param to jenkins so tom can trigger builds for libntech PRs.

olehermanse:
@vpodzime I think you will like the second commit here ;) 

https://github.com/cfengine/libntech/pull/8/commits/5f392545f0829616f33dd9bbcb918e2f79ee75f2

PR #9, opened 2019-09-10
Reimplemented isprint to fix issues on HPUX

olehermanse:


olehermanse:
Running tests here: https://github.com/cfengine/core/pull/3826

libutils/string_lib.h
```diff
@@ -81,6 +81,11 @@ char *NULLStringToEmpty(char *str);
 
 bool StringIsNumeric(const char *name);
 bool StringIsPrintable(const char *name);
+// isprint has weird behavior on HPUX, here is a replacement:
```
amousset: Out of curiosity, what kind of weird behavior? :open_mouth:

olehermanse: @amousset I can clarify some more in commit message / comments.

The commit you linked is the same everywhere, since isprint takes a signed int, it will implicitly cast the char to int, extending the sign bit. So always cast to unsigned char first.

The issue on HPUX seems to be that isprint returns true for some characters above 127.

> These functions classify character-coded integer values according to the rules of the coded character set identified by the last successful call to setlocale() (see setlocale(3C)). Each function is a predicate returning nonzero for true, zero for false. If setlocale() has not been called successfully, characters are classified according to the rules of the default ASCII 7-bit coded character set (see setlocale(3C)). https://docstore.mik.ua/manuals/hp-ux/en/B2355-60130/ctype.3C.html

https://docstore.mik.ua/manuals/hp-ux/en/B2355-60130/ctype.3C.html

So, I guess depending on locale they will identify some more characters as printable. That's useful if you want to print directly to terminal but in this case we want predictable behavior across platforms.

Our version of isprint will use the printable ascii range, which seems to be how it works everywhere else.

olehermanse: @amousset I expanded the comment and amended the commit:

```
/**
 * @brief Check if a char is "printable", replacement for isprint
 *
 * isprint takes a (signed) int, so if you send in a (signed) char there is an
 * implicit cast. This can be problematic, since 0xF0 becomes 0xFFFFF0.
 *
 * Additionally, at least on HPUX, isprint returns true for some values above
 * 127, depending on locale.
 *
 * Thus, this replacement works the same everywhere, and can be used when you
 * want predictable results, even though it will return false for some values
 * which might be possible to print in your environment / locale.
 *
 * @param c [in] Character to evaluate
 * @return True if c is part of the printable ascii range
 */
static inline bool CharIsPrintableAscii(const char c)
{
    return (c >= ' ' && c <= '~');
}
```

amousset: Thanks for the explanation!

libutils/string_lib.h
```diff
@@ -81,6 +81,27 @@ char *NULLStringToEmpty(char *str);
...
+ * 127, depending on locale.
+ *
+ * Thus, this replacement works the same everywhere, and can be used when you
+ * want predictable results, even though it will return false for some values
+ * which might be possible to print in your environment / locale.
+ *
+ * @param c [in] Character to evaluate
+ * @return True if c is part of the printable ascii range
+ */
+static inline bool CharIsPrintableAscii(const char c)
```
vpodzime: Hmm, code in a header file?

karlhto: Yeah, this is how you "force" this function to be inlined in every function where you include the header file. Since the function is so small I would argue that it makes sense to do it this way.

PR #12, opened 2019-09-24
Automake: Added CORE_ variables to libutils

olehermanse:


olehermanse:
Being tested here: https://github.com/cfengine/core/pull/3859

PR #17, opened 2019-10-04
Split Stack into 2 versions, one thread-safe, one not

olehermanse:


olehermanse:
> 7 for Pointer argument is dereferenced without checking for NULL

Query looks bugged, I'll have to take a look.

olehermanse:
> I don't really like the code duplication. To me the difference can be done by a couple of ifs or #ifdefs.

@vpodzime 

- This is how we do it for queue, already.
- I don't think generating these data structures using macros is worth it. Ref. StringMap confusion.
- I'd like to add different methods to (non-threaded) stack, most notably, referencing the top element, without popping it.

vpodzime:
> I'd like to add different methods to (non-threaded) stack, most notably, referencing the top element, without popping it.

Why is `StackTop()` not useful for the *ThreadedStack*?

And regarding the *Queue* -- I was suggesting the same there. I just don't like that any fixes and improvements will have to go to two places.

olehermanse:
> Why is StackTop() not useful for the ThreadedStack?

Because it would allow multiple threads to access the same data, without locking. One thread could have a reference to the top element, while another is popping and freeing that same element.

> And regarding the Queue -- I was suggesting the same there. I just don't like that any fixes and improvements will have to go to two places.

Unfortunately, I don't think there are nice ways of doing this in C. I tried converting the code to use preprocessor tricks to reuse the same base code yesterday `#define StackType ThreadedStack` `#include <stack_base.c>` etc. It was not pretty, and I gave up after a while.

```
bool StackType ## IsEmpty(StackType *stack);
```

There are 2 other options:

- Defining a DEFINE / DECLARE macro, exactly like StringMap is created. I think this is hard to search for and hard to read. This is, arguably, the only one which scales to many data structures for different types.
- Refactoring `ThreadedStack` to be a struct of `Stack` + `lock`. Make all the `Threaded` methods be wrappers that do `ThreadLock` + `StackMethod` + `ThreadUnlock`.

vpodzime:
>     * Refactoring `ThreadedStack` to be a struct of `Stack` + `lock`. Make all the `Threaded` methods be wrappers that do `ThreadLock` + `StackMethod` + `ThreadUnlock`

Can't we make `Stack` have the `lock` field and only use it if it is `!= NULL`? Then we would just have two constructors, one initializing the lock and another one just setting it to `NULL`. (This is what I wanted for the `Queue`, btw).

olehermanse:
> Can't we make Stack have the lock field and only use it if it is != NULL? Then we would just have two constructors, one initializing the lock and another one just setting it to NULL. (This is what I wanted for the Queue, btw).

@vpodzime We could, but I much prefer to have the separate type and type checking.

PR #21, opened 2019-10-11
Cleaned up includes to rely on system / standard headers instead of config.h

olehermanse:
Working on new project here:

https://github.com/olehermanse/cf-syntax


olehermanse:
Being tested here: https://github.com/cfengine/core/pull/3900

olehermanse:
> _stdbool.h_ **IS** already being included everywhere, right? Looks good to me otherwise!

@vpodzime Ah, that part of the commit message is outdated now, removed.

PR #22, opened 2019-10-11
hash.h: ensure type 'bool' has been declared

tdlee:
"hash.h" uses type 'bool' in some declarations. When various files in CFEngine include "hash.h" the 'bool' type might not have been included.  This commit ensures the validity of 'bool'.

### DIFF
--- a/libutils/hash.h
+++ b/libutils/hash.h
@@ -29,6 +29,7 @@
   @brief Hash implementations
   */
 
+#include <stdbool.h>
 #include <openssl/rsa.h>
 #include <openssl/evp.h>
 


cf-bottom:
Thank you for submitting a pull request! Maybe @olehermanse can review this?

olehermanse:
Hello, thank you for opening a pull request. I am already looking at this:

https://github.com/cfengine/core/pull/3888/files#diff-e2e414c518c22210534fc9de14b1acd5R28

However, as you can see, here: 

https://github.com/cfengine/libntech/blob/master/libutils/bool.h#L30-L44

We already have some macros for detecting whether `stdbool.h` exists, presumably because it's missing on some of the platforms we build on. So, I'll want to remove `bool.h` and the related macros in favor of `stdbool.h` everywhere, but I need some time to test that this works across our platforms.

I'll follow up on this when I have time.

PR #23, opened 2019-10-14
Include platform.h before pthread.h

olehermanse:
When compiling on Debian 6 (in core) errors were encountered,
because, on this platform, it's important to set some
OPEN_SOURCE macro before including `pthread.h`:

https://stackoverflow.com/questions/18375527/why-is-gnu-source-macro-required-for-pthread-mutexattr-settype-while-it-is-in

### DIFF
--- a/libutils/threaded_deque.c
+++ b/libutils/threaded_deque.c
@@ -22,11 +22,12 @@
   included file COSL.txt.
 */
 
+#include <platform.h>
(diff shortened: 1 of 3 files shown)


olehermanse:
Tested, here:

https://github.com/cfengine/core/pull/3888#issuecomment-541699102

Fixed the errors while compiling.

PR #24, opened 2019-10-16
Remove bool.h and use stdbool.h everywhere

olehermanse:
This header is problematic because it assumes a config.h
which defines HAVE_STDBOOL_H before including it.

Let's try to just use stdbool.h directly, and see if it works on
all our platforms.


olehermanse:
Being tested here: https://github.com/cfengine/core/pull/3899

PR #27, opened 2019-10-18
File locking

vpodzime:
Merge together:
https://github.com/cfengine/core/pull/3880
https://github.com/cfengine/enterprise/pull/556
https://github.com/cfengine/libntech/pull/27

### DIFF
--- a/libutils/file_lib.c
+++ b/libutils/file_lib.c
@@ -1509,3 +1509,145 @@ const char* GetRelocatedProcdirRoot()
 
     return procdir;
 }
+
+
+#if !defined(__MINGW32__)
+
+static int LockFile(int fd, short int lock_type, bool wait)
+{
+    struct flock lock = {
+        .l_type = lock_type,
+        .l_whence = SEEK_SET,
+        .l_start = 0, /* start of the region to which the lock applies */
+        .l_len = 0    /* till EOF */
+    };
+
+    if (wait)
+    {
+        while (fcntl(fd, F_SETLKW, &lock) == -1)
+        {
+            if (errno != EINTR)
+            {
+                return -1;
+            }
+        }
+        return 0;
+    }
+    else
+    {
+        int ret = fcntl(fd, F_SETLK, &lock);
+        /* make sure we only return -1 or 0 like block above */
+        return ret == -1 ? -1 : 0;
+    }
+}
+
+int ExclusiveLockFile(int fd, bool wait)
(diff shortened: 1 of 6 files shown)


libutils/file_lib.c
```diff
@@ -1563,6 +1563,28 @@ int ExclusiveFileLock(FileLock *lock, bool wait)
     return LockFD(lock->fd, F_WRLCK, wait);
 }
 
+int ExclusiveFileLockPath(FileLock *lock, const char *fpath, bool wait)
+{
+    assert(lock != NULL);
+    assert(lock->fd < 0);
```
olehermanse: There's quite a few places where you assign `lock->fd` to -1 or compare it to 0. I would add a comment or use the macro (`EMPTY_FILE_LOCK`) to make it more obvious what this means.

libutils/logging.c
```diff
@@ -273,7 +275,34 @@ const char *GetErrorStr(void)
...
+#else
+
+const char *GetErrorStrFromCode(int error_code)
+{
+    static char errbuf[CF_BUFSIZE];
+    int len;
+
+    memset(errbuf, 0, sizeof(errbuf));
+
+    if (FormatMessage(FORMAT_MESSAGE_FROM_SYSTEM, NULL, error_code,
```
olehermanse: ❤️ 💟 ❤️ 💟 ❤️ 💟

libutils/logging.c
```diff
@@ -273,7 +275,34 @@ const char *GetErrorStr(void)
...
+                  MAKELANGID(LANG_ENGLISH, SUBLANG_ENGLISH_US), errbuf, CF_BUFSIZE, NULL))
+    {
+        // remove CRLF from end
+        len = strlen(errbuf);
+        errbuf[len - 2] = errbuf[len - 1] = '\0';
```
olehermanse: What if `len` is `0`?

libutils/file_lib.c
vpodzime: It's an invalid file descriptor, what's pretty clear :wink: Yeah, I'll add a note that it's used as an indicator if the lock is initialized or not.

libutils/logging.c
vpodzime: No idea, I'm not touching this code :smile:

vpodzime: (just moved it around)

PR #29, opened 2019-10-21
Fix issues in the file locking implementation for Windows

vpodzime:
Copy-pasta failures and a missing include.

### DIFF
--- a/libutils/file_lib.c
+++ b/libutils/file_lib.c
@@ -36,6 +36,8 @@
 
 #ifndef __MINGW32__
 #include <glob.h>
+#else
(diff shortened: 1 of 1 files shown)


libutils/file_lib.c
```diff
@@ -36,6 +36,7 @@
 
 #ifndef __MINGW32__
 #include <glob.h>
+#include <Windows.h>
```
olehermanse: Should be in an else, i.e. not for mingw32

vpodzime: Fixed.

PR #35, opened 2019-11-01
Added StringCloseHole() from core

olehermanse:
This was a static function in core, called CloseStringHole().
Added asserts, tests, and a doxygen comment.

### DIFF
--- a/libutils/string_lib.c
+++ b/libutils/string_lib.c
@@ -1187,6 +1187,19 @@ size_t TrimCSVLineCRLFStrict(char *const data)
     return length;
 }
 
+void StringCloseHole(char *s, const int start, const int end)
+{
+    assert(s != NULL);
+    assert(0 <= start && start <= end && end <= strlen(s));
+    assert((end - start) <= strlen(s));
+
+    if (end > start)
+    {
+        memmove(s + start, s + end,
+                /* The 1+ ensures we copy the final '\0' */
+                strlen(s + end) + 1);
+    }
+}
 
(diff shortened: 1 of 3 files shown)


libutils/string_lib.c
```diff
@@ -1187,6 +1187,19 @@ size_t TrimCSVLineCRLFStrict(char *const data)
     return length;
 }
 
+void StringCloseHole(char *s, int start, int end)
```
karlhto: `const` for `start` and `end`

libutils/string_lib.c
```diff
@@ -1187,6 +1187,19 @@ size_t TrimCSVLineCRLFStrict(char *const data)
...
+{
+    assert(s != NULL);
+    assert(0 <= start && start <= end && end <= strlen(s));
+    assert((end - start) <= strlen(s));
+
+    if (end > start)
+    {
+        memmove(s + start, s + end,
+                /* The 1+ ensures we copy the final '\0' */
+                1 + strlen(s + end));
```
karlhto: Just a preference thing: maybe `strlen(...) + 1` instead, since the `NUL`-byte is at the end?

olehermanse: Didn't really want to change anything, just move the function. Added assertions, but nothing else.

olehermanse: Fixed.

libutils/string_lib.c
olehermanse: Added.

PR #38, opened 2019-11-28
Refactor json-utils, put `JsonParseAnyFile` in json.h header

karlhto:
Refactoring includes:
  - Put `FileType` in `json-utils` and make it an enum instead
    (previously just returned a string)
  - Added utility function to get a string from the new enums
  - Added JsonParseAnyFile to `json.h`. It wasn't static, and is
    actually (slightly) useful outside `json.c`

Note that this needs to be merged with a libpromises patch that replaces
the use of `JsonReadDataFile`

Ticket: None
Changelog: None

I was unsure of whether to split this into several commits or not. It should be fairly easy to review I think.

### DIFF
--- a/libutils/json-utils.c
+++ b/libutils/json-utils.c
@@ -302,13 +302,12 @@ bool JsonParseCsvFile(const char *input_path, size_t size_max, JsonElement **jso
 }
 
 JsonElement *JsonReadDataFile(const char *log_identifier, const char *input_path,
-                          const char *requested_mode, size_t size_max)
+                              const DataFileType requested_mode, size_t size_max)
 {
     const char *myname = log_identifier ? log_identifier : "JsonReadDataFile";
-
-    bool env_mode = (strcmp("ENV", requested_mode) == 0);
-    bool csv_mode = (strcmp("CSV", requested_mode) == 0);
-    //size_t size_max = 50 * (1024 * 1024);
+    bool env_mode = (requested_mode == DATAFILETYPE_ENV);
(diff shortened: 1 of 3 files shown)


libutils/json-utils.c
```diff
@@ -329,34 +327,31 @@ JsonElement *JsonReadDataFile(const char *log_identifier, const char *input_path
...
     JsonParseError res;
-    if (yaml_mode)
-    {
-        res = JsonParseYamlFile(input_path, size_max, &json);
-    }
-    else
-    {
-        res = JsonParseFile(input_path, size_max, &json);
-    }
+    bool yaml_mode = (requested_mode == DATAFILETYPE_YAML);
```
vpodzime: I'd do this together with the other comparisons of `requested_mode`.

karlhto:
@vpodzime @olehermanse Just rebased to master, no additional changes. : )

libutils/json-utils.c
```diff
@@ -366,3 +361,64 @@ JsonElement *JsonReadDataFile(const char *log_identifier, const char *input_path
...
+    case DATAFILETYPE_CSV:
+        return "CSV";
+    case DATAFILETYPE_YAML:
+        return "YAML";
+    case DATAFILETYPE_ENV:
+        return "ENV";
+    case DATAFILETYPE_JSON:
+        return "JSON";
+    default:
+        return "unknown";
```
olehermanse: +1. Thanks for not using an array of strings here.

karlhto:
Again, just rebased to master. No additional changes.

Travis should now work correctly as per the changes @olehermanse made.

PR #40, opened 2019-12-16
Fix file lock test

vpodzime:


vpodzime:
good news from AIX 5.3:
```
14:31:53 PASS: file_lock_test
```

PR #46, opened 2020-01-30
ENT-5313: SeqStringWriteFileStream() no longer closes FILE stream

olehermanse:
This was not really intended, and it is very useful to be able to keep writing to the file afterwards.

### DIFF
--- a/libutils/string_sequence.c
+++ b/libutils/string_sequence.c
@@ -176,7 +176,7 @@ bool SeqStringWriteFileStream(Seq *seq, FILE *file)
     assert(w != NULL);
 
     bool success = SeqStringWrite(seq, w);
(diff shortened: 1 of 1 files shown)


libutils/string_sequence.c
```diff
@@ -187,7 +187,9 @@ bool SeqStringWriteFile(Seq *seq, const char *file)
     {
         return false;
     }
-    return SeqStringWriteFileStream(seq, f);
+    bool success = SeqStringWriteFileStream(seq, f);
+    fclose(f);
```
vpodzime: I guess we don't care about any write issues happening here?

olehermanse: Fixed.

PR #61, opened 2020-02-28
Log messages about broken Mustache templates are now errors

olehermanse:
Previously, they were warnings.

According to core/CONTRIBUTING.md:

* `LOG_LEVEL_ERR`: Promise failed or other errors that are definitely
  considered bad / not normal.
* `LOG_LEVEL_WARNING`: Something unusual happened that the user should
  investigate. Should be severe enough to warrant investigating further,
  but not as severe as a definitive error/bug.

I would argue that if you have a broken template, that definitely
warrants an error. It means there is a bug in your template, or
whatever is generating the template. Looking at `git blame` this
has been a warning since mustache was introduced, and there is no
specific reasoning behind it.

### DIFF
--- a/libutils/mustache.c
+++ b/libutils/mustache.c
@@ -274,7 +274,7 @@ static Mustache NextTag(const char *input,
         const char *escape_end = strstr(ret.content, extra_end);
         if (!escape_end || strncmp(escape_end + 1, delim_end, delim_end_len) != 0)
         {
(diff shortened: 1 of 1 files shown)


olehermanse:
@cf-bottom jenkins

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/4715/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/4715/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-4715/

PR #66, opened 2020-04-03
CFE-3285 cfengine for termux environment

craigcomstock:
merge together
https://github.com/cfengine/libntech/pull/66
https://github.com/cfengine/core/pull/4132

### DIFF
--- a/libutils/file_lib.c
+++ b/libutils/file_lib.c
@@ -562,7 +562,7 @@ int safe_open_create_perms(
 #ifdef __MINGW32__
     // Windows gets off easy. No symlinks there.
     return open(pathname, flags, create_perms);
-#elif __TERMUX__
+#elif defined(__ANDROID__)
     // if effective user is not root then don't try to open
     // all paths from '/' up, might not have permissions.
     uid_t p_euid = geteuid();
@@ -570,7 +570,7 @@ int safe_open_create_perms(
     {
         return open(pathname, flags, create_perms);
     }
-#else // !__MINGW32__ and !__TERMUX__
+#else // !__MINGW32__ and !__ANDROID__
     const size_t path_bufsize = strlen(pathname) + 1;
     char path[path_bufsize];
     const size_t res_len = StringCopy(pathname, path, path_bufsize);
--- a/libutils/file_lib.c
+++ b/libutils/file_lib.c
@@ -562,8 +562,15 @@ int safe_open_create_perms(
 #ifdef __MINGW32__
     // Windows gets off easy. No symlinks there.
     return open(pathname, flags, create_perms);
-#else // !__MINGW32__
-
+#elif __TERMUX__
+    // if effective user is not root then don't try to open
+    // all paths from '/' up, might not have permissions.
+    uid_t p_euid = geteuid();
+    if (p_euid != 0)
+    {
+        return open(pathname, flags, create_perms);
+    }
+#else // !__MINGW32__ and !__TERMUX__
     const size_t path_bufsize = strlen(pathname) + 1;
     char path[path_bufsize];
     const size_t res_len = StringCopy(pathname, path, path_bufsize);
--- a/libutils/known_dirs.c
+++ b/libutils/known_dirs.c
@@ -98,6 +98,7 @@ const char *GetDefault##FUNC##Dir(void)                             \
     return GetDefaultDir_helper(STATIC##dir, GLOBAL, FOLDER);       \
 }
 
+#if !defined(__CYGWIN__) && !defined(__ANDROID__)
 GET_DEFAULT_DIRECTORY_DEFINE(Work, work, WORKDIR, NULL)
 GET_DEFAULT_DIRECTORY_DEFINE(Bin, bin, BINDIR, "bin")
 GET_DEFAULT_DIRECTORY_DEFINE(Log, log, LOGDIR, "log")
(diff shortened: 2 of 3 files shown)


tests/unit/file_lock_test.c
```diff
@@ -49,6 +49,7 @@ static void clear_tempfiles()
 int main()
 {
     atexit(&clear_tempfiles);
+    printf("mkdir TEMPDIR = '%s'\n", TEMP_DIR);
```
craigcomstock, line 52: remove this debug printf.

craigcomstock:
Ok. file_lib_test is not happy. Will work on it.

craigcomstock:
@cf-bottom jenkins with exotics

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/5354/)

(with exotics)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/5354/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-5354/

craigcomstock:
@vpodzime for your weekend fun, please review this and https://github.com/cfengine/core/pull/4132, thanks!

libutils/known_dirs.c
```diff
@@ -106,6 +105,8 @@ GET_DEFAULT_DIRECTORY_DEFINE(Master, master, MASTERDIR, "masterfiles")
 GET_DEFAULT_DIRECTORY_DEFINE(Input, input, INPUTDIR, "inputs")
 GET_DEFAULT_DIRECTORY_DEFINE(State, state, STATEDIR, "state")
 
+#endif
```
vpodzime, line 108: This seems to be breaking mingw builds? https://ci.cfengine.com/job/testing-pr/label=PACKAGES_i386_mingw/9618/

libutils/platform.h
```diff
@@ -178,6 +178,9 @@ struct utsname
 #endif
 
 #include <signal.h>
+#ifdef __ANDROID__
+#include <android/log.h>
+#endif
```
vpodzime, line 183: Should this go right next to the `ifdef` below? I take it it defines the same? Or maybe it should go to `logging.h`?

craigcomstock: Checking why I needed this at all. Then will move and comment appropriately.

libutils/known_dirs.c
craigcomstock: Yes. The logic is a bit gnarly. Working on it.

libutils/file_lib.c
```diff
@@ -562,8 +562,15 @@ int safe_open_create_perms(
 #ifdef __MINGW32__
     // Windows gets off easy. No symlinks there.
     return open(pathname, flags, create_perms);
-#else // !__MINGW32__
-
+#elif __TERMUX__
```
craigcomstock, line 565: should probably change to `__ANDROID__` for on-device build-ability.

libutils/platform.h
craigcomstock: Not sure why I had this. It doesn't seem to be needed so I removed it.

craigcomstock:
will squash commits tomorrow

craigcomstock:
@vpodzime commits squashed, let's try and wrap this up yeah? :) I think the core PR is green enough.

libutils/file_lib.c
vpodzime: This should be `#elif defined(__TERMUX__)` AFAICT.

libutils/known_dirs.c
```diff
@@ -98,13 +98,16 @@ const char *GetDefault##FUNC##Dir(void)                             \
     return GetDefaultDir_helper(STATIC##dir, GLOBAL, FOLDER);       \
 }
 
+#if !defined(__CYGWIN__) && !defined(__ANDROID__)
```
vpodzime, line 101: Why `__CYGWIN__` and not `__MINGW32__`?

craigcomstock: Because line 31

```
#if defined(__CYGWIN__) || defined(__ANDROID__)
```

defines and uses GET_DEFAULT_DIRECTORY_DEFINE to define such functions as GetDefaultWorkDir().

But MINGW32 needs this block which uses the OTHER GET_DEFAULT_DIRECTORY_DEFINE() macro which uses the GetDefaultDir_helper() function, in MINGW32 case that's in enterprise. :|

So I would guess libntech won't build with `__MINGW32__` and maybe we should handle this better? Maybe move some or all of this file back to core to live along-side enterprise better?

libutils/file_lib.c
craigcomstock: so long as `__TERMUX__`, if present, evaluates to non-zero, which I think it does, then `#elif` is "ok" but I don't disagree that `#elif defined(__TERMUX__)` is more clear and resilient.

craigcomstock: https://gcc.gnu.org/onlinedocs/cpp/If.html#If

```
Identifiers that are not macros, which are all considered to be the number zero. This allows you to write #if MACRO instead of #ifdef MACRO, if you know that MACRO, when defined, will always have a nonzero value. 
```

libutils/known_dirs.c
vpodzime: > So I would guess libntech won't build with `__MINGW32__` and maybe we should handle this better? Maybe move some or all of this file back to core to live along-side enterprise better?

Yeah, it is definitely not ideal, but that's out of scope of this PR.

vpodzime:
Please squash the last commit where it belongs or modify it to be a proper commit on its own.

craigcomstock:
@vpodzime thanks for the review. I think this is in order, please check and merge at will. :)
I don't think we need to back port, can leave in master and for 3.17.

PR #69, opened 2020-05-12
Added system group name defines

craigcomstock:
This is supporting a core change https://github.com/cfengine/core/pull/4178

Ticket: ENT-5986
Changelog: Title


cf-bottom:
Thanks for submitting a PR! Maybe @vpodzime can review this?

PR #71, opened 2020-05-25
C preprocessor: 'elseif' is illegal; should be 'elif'

tdlee:


tdlee:
Shouldn't CI have detected this fault when it was introduced (which was about two weeks ago)?

olehermanse:
@tdlee Good catch.

> Shouldn't CI have detected this fault when it was introduced (which was about two weeks ago)?

FreeBSD is not an officially supported platform and we have no CI using FreeBSD. Because of how C preprocessor macros work, this error will only show up on FreeBSD, on all other platforms `__FreeBSD__` is not defined and the problematic macro is part of the skipped code until a matching `#elif` or `#endif`.

olehermanse:
CFEngine core is being updated here: https://github.com/cfengine/core/pull/4197

(Once tests are done).

craigcomstock:
Sorry about the breakage.

PR #75, opened 2020-06-05
misc files: Add explicit checks for NULL before pointer dereferences

lluiscampos:
Gets the rid of a bunch of LGTM alerts. Minor style and consistency
fixes along the way.

### DIFF
--- a/libutils/buffer.c
+++ b/libutils/buffer.c
@@ -50,6 +50,7 @@ Buffer *BufferNew(void)
 
 static void ExpandIfNeeded(Buffer *buffer, unsigned int needed)
 {
+    assert(buffer != NULL);
     if (needed >= buffer->capacity)
     {
         size_t new_capacity = UpperPowerOfTwo(needed + 1);
@@ -68,7 +69,7 @@ Buffer* BufferNewFrom(const char *data, unsigned int length)
 
 void BufferDestroy(Buffer *buffer)
 {
-    if (buffer)
+    if (buffer != NULL)
     {
         free(buffer->buffer);
         free(buffer);
@@ -77,6 +78,7 @@ void BufferDestroy(Buffer *buffer)
 
 char *BufferClose(Buffer *buffer)
 {
+    assert(buffer != NULL);
     char *detached = buffer->buffer;
     free(buffer);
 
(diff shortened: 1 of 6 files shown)


libutils/buffer.c
```diff
@@ -447,26 +454,23 @@ const char* BufferSearchAndReplace(Buffer *buffer, const char *pattern, const ch
...
 {
-    assert(buffer);
+    assert(buffer != NULL);
     buffer->used = 0;
     buffer->buffer[0] = '\0';
 }
 
 unsigned int BufferSize(const Buffer *buffer)
 {
-    assert(buffer);
```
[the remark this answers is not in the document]
lluiscampos: Roger that.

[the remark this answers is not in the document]
lluiscampos: Fixed in the last push.

libutils/sequence.c
```diff
@@ -58,16 +59,19 @@ static void DestroyRange(Seq *seq, size_t start, size_t end)
...
     {
-        DestroyRange(seq, 0, seq->length - 1);
+        if (seq->length > 0)
+        {
+            DestroyRange(seq, 0, seq->length - 1);
+        }
+        SeqSoftDestroy(seq);
     }
-    SeqSoftDestroy(seq);
 }
```
olehermanse, line 70: Not a big deal, but I think this is more clear with an early return, instead of nested ifs.

**Original:**

```C
void SeqDestroy(Seq *seq)
{
    if (seq && seq->length > 0)
    {
        DestroyRange(seq, 0, seq->length - 1);
    }
    SeqSoftDestroy(seq);
}
```

**Your refactor:**


```C
void SeqDestroy(Seq *seq)
{
    if (seq != NULL)
    {
        if (seq->length > 0)
        {
            DestroyRange(seq, 0, seq->length - 1);
        }
        SeqSoftDestroy(seq);
    }
}
```

**My suggestion:**

```C
void SeqDestroy(Seq *seq)
{
    if (seq == NULL)
    {
        return;
    }

    if (seq->length > 0)
    {
        DestroyRange(seq, 0, seq->length - 1);
    }
    SeqSoftDestroy(seq);
}
```

Just an FYI.

olehermanse: More interesting reading: https://refactoring.guru/replace-nested-conditional-with-guard-clauses

lluiscampos: Thanks for the advise. The early return recommendation is something I am aware of, I'll have it in mind next time

PR #80, opened 2020-06-22
Cleaned up IP address comparison functions

olehermanse:
More cases of functions and variables which should be bool.
AFAICT the -1 return value was unreachable anyway. (And it was
treated as bool in the calling location).

Also refactored IPAddressCompareLess() to follow code style,
fixed indentation, added assertions and comments to make it
easier to understand.

Reported by LGTM:
https://lgtm.com/rules/1508604166004/

### DIFF
--- a/libutils/ip_address.c
+++ b/libutils/ip_address.c
@@ -1187,79 +1187,85 @@ int IPAddressIsEqual(IPAddress *a, IPAddress *b)
 /*
  * Sorting comparison for IPV4 addresses
  */
-static int IPV4CompareLess(struct IPV4Address *a, struct IPV4Address *b)
+static bool IPV4CompareLess(struct IPV4Address *a, struct IPV4Address *b)
 {
     int i = 0;
     for (i = 0; i < 4; ++i)
     {
         if (a->octets[i] > b->octets[i])
         {
-            return 0;
+            return false;
         }
         else if (a->octets[i] < b->octets[i])
         {
-            return 1;
+            return true;
         }
     }
 
-    return 0;
+    return false;
 }
 
 /*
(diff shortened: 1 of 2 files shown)


libutils/ip_address.c
```diff
@@ -1235,31 +1235,34 @@ int IPAddressCompareLess(IPAddress *a, IPAddress *b)
      */
      if (!a || !b)
      {
-         return 1;
+         return true;
      }
 
+    assert(a->type == IP_ADDRESS_TYPE_IPV4 || a->type == IP_ADDRESS_TYPE_IPV6);
```
vpodzime, line 1241: Bad indentation.

libutils/ip_address.h
```diff
@@ -91,6 +91,6 @@ bool IPAddressIsIPAddress(Buffer *source, IPAddress **address);
   @param b IP address of the second object.
   @return 1 if a < b, and 0 otherwise.
```
vpodzime, line 92: Needs to be updated.

libutils/ip_address.c
olehermanse: Most of the function is wrongly indented at 5 spaces, this is correct, 4 spaces.

libutils/ip_address.h
olehermanse: Thanks, fixed.

libutils/ip_address.c
olehermanse: Fixed the rest of the indentation in a second commit.

olehermanse: Refactored whole function and squashed into 1 commit.

PR #82, opened 2020-08-25
Add a new function for reading lines from a file into a sequence

vpodzime:
Working with individual lines from a file is a common pattern and
sequence is a great data structure for storing such data.

Ticket: CFE-2945
Changelog: None


olehermanse (approved):
LGTM

PR #85, opened 2020-09-08
CFE-3409 definitions.h: Added constant buffer sizes for DB names and table names

olehermanse:
Placed them here because they have a loose relationship to
CF_MAXVARSIZE. Just constants, shouldn't hurt anyone.


vpodzime (approved):
ACK

PR #86, opened 2020-10-05
Added StringContains() and StringContainsChar()

olehermanse:
Inlined functions for better readability.

### DIFF
--- a/libutils/string_lib.h
+++ b/libutils/string_lib.h
@@ -102,6 +102,16 @@ static inline bool CharIsPrintableAscii(const char c)
 }
 bool EmptyString(const char *s);
 
+static inline bool StringContains(const char *const haystack, const char *const needle)
+{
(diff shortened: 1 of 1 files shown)


libutils/string_lib.h
```diff
@@ -102,6 +102,16 @@ static inline bool CharIsPrintableAscii(const char c)
 }
 bool EmptyString(const char *s);
 
+static inline bool StringContains(const char *const a, const char *const b)
```
vpodzime, line 105: `needle` and `hay` would be better names for the parameters.

vpodzime (commented):
Looks good otherwise.

libutils/string_lib.h
olehermanse: Done.

vpodzime (approved):
ACK

PR #98, opened 2020-11-24
Add ARM64 job in Travis-CI

odidev:
- .travis.yml : Added Arm64 jobs along with the installation of package require for arm64
 - libutils/[buffer.c, string_lib.c]: Fixed strcmp() issue for arm64
   - strcmp() has different behavior in arm64 and amd64. At ARM64, strcmp(char* a, char* b) returns the value other than 1 or -1, as per given char buffer inputs.
   - While asserting buf_size in libutils/string_lib.c in line 65, added typecasting as in ARM64 'char' is considered as 'unsigned char' whereas AMD64 treats 'char' as 'signed char'

Signed-off-by: odidev <odidev@puresoftware.com>

### DIFF
--- a/libutils/string_lib.c
+++ b/libutils/string_lib.c
@@ -231,7 +231,14 @@ int StringSafeCompare(const char *const a, const char *const b)
     }
     if (a != NULL && b != NULL)
     {
-        return strcmp(a, b);
+        // Adding this as strcmp gives difference of the buffer values in aarch64
+        // Whereas it gives 1, 0 or -1 in other platforms accordingly.
+        int compare_result = strcmp(a, b);
+        if (compare_result != 0)
+        {
+            compare_result = compare_result / abs(compare_result);
+        }
+        return compare_result;
     }
 
     // Weird edge cases where one is NULL:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: c
-dist: xenial
+dist: bionic
 
 addons:
   artifacts:
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,8 +38,21 @@ matrix:
     - os: osx
       env: JOB_TYPE=compile_and_unit_test COVERAGE=no
       sudo: false
+    - arch: arm64
+      os: linux
+      env: JOB_TYPE=compile_and_unit_test_asan
+    - arch: arm64
+      os: linux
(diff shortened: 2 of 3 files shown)


libutils/string_lib.c
```diff
@@ -62,7 +62,7 @@ size_t StringCopy(const char *const from, char *const to, const size_t buf_size)
     assert(from != NULL);
     assert(to != NULL);
     assert(from != to);
-    assert(buf_size >= 0);
+    assert((signed)buf_size >= 0);
```
vpodzime, line 65: `buf_size` is `size_t`. There should be no problem comparing that to `0`. This looks wrong.

vpodzime (commented):
One of the changes doesn't look correct to me. Also, could you please wrap the long lines in the commit message? And could you please split the change of the Ubuntu version used in Travis into a separate commit? It would make searching for the change in git history much easier.

Thanks for adding this!

odidev:
Thanks for the suggestion, 

I have added another commit for upgrading the distro version and tried to wrap the comments.

libutils/string_lib.c
odidev: I have reverted the changes. 

These changes were required if we build the package over gcc but we are using clang in travis for **“env: JOB_TYPE=compile_and_unit_test COVERAGE=no”** so these changes are not required.

olehermanse: You have not reverted the change, it's still here. I think you're doing this because of a warning for comparison always true, because of the sign of the type.

We typically disable this by adding `-Wno-sign-compare`.

odidev: Yes, my intention for change was the same as you described.
I have reverted that type-conversion now, as that is not required with clang being used in our travis-ci build.

vpodzime (approved):
Looks good to me now.

vpodzime:
@cf-bottom run this through jenkins, please

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/5961/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/5961/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-5961/

vpodzime:
> Build Status

Only unrelated failures. :+1:

olehermanse (approved):

PR #102, opened 2020-12-17
Don't redefine le32toh and htole32

olehermanse:
Necessary to try compiling with `-std=c11` in gcc, since
they are already defined then.


larsewi (approved):
Awesome !

vpodzime (approved):
ACK

PR #103, opened 2020-12-17
Removed line split to ease debugging

olehermanse:
This made it more annoying to debug for me, since
the compiler would point me to the last (shared) line,
which doesn't have the relevant information (the return type).


larsewi (commented):
Should we do the same with `setnetgrent` above for consistency?

vpodzime (approved):
ACK

PR #108, opened 2021-01-09
CFE-3540: GitHub Actions: Added MacOS Unit Tests

olehermanse:
At first I was adding it to the same workflow as Ubuntu, but
this became ugly, since there are many differences. A separate
workflow (yml) makes it much cleaner.

Running with ASAN on MacOS requires some more work, but this gets us halfway there.

### DIFF
--- a/.github/workflows/asan_unit_tests.yml
+++ b/.github/workflows/asan_unit_tests.yml
@@ -1,4 +1,4 @@
-name: Build and unit test
+name: ASAN Unit Tests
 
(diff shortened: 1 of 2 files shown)


olehermanse:
Identical to: https://github.com/cfengine/core/pull/4445

nickanderson (approved):

larsewi (approved):

vpodzime (approved):
ACK

PR #109, opened 2021-01-10
CFE-3415: Fixed some sign-compare warnings

larsewi:
Ticket: CFE-3415
Changelog: None

### DIFF
--- a/libutils/buffer.c
+++ b/libutils/buffer.c
@@ -32,7 +32,7 @@
 #endif
 #include <string_lib.h>
 
-Buffer *BufferNewWithCapacity(unsigned int initial_capacity)
+Buffer *BufferNewWithCapacity(size_t initial_capacity)
 {
     Buffer *buffer = xmalloc(sizeof(Buffer));
 
@@ -50,7 +50,7 @@ Buffer *BufferNew(void)
     return BufferNewWithCapacity(DEFAULT_BUFFER_CAPACITY);
 }
 
-static void ExpandIfNeeded(Buffer *buffer, unsigned int needed)
+static void ExpandIfNeeded(Buffer *buffer, size_t needed)
 {
     assert(buffer != NULL);
     if (needed >= buffer->capacity)
@@ -61,7 +61,7 @@ static void ExpandIfNeeded(Buffer *buffer, unsigned int needed)
     }
 }
 
-Buffer* BufferNewFrom(const char *data, unsigned int length)
+Buffer* BufferNewFrom(const char *data, size_t length)
 {
     Buffer *buffer = BufferNewWithCapacity(length + 1);
     BufferAppend(buffer, data, length);
@@ -126,7 +126,7 @@ int BufferCompare(const Buffer *buffer1, const Buffer *buffer2)
              * BUFFER_BEHAVIOR_BYTEARRAY
              * Byte by byte comparison
              */
-            unsigned int i = 0;
+            size_t i = 0;
             if (buffer1->used < buffer2->used)
             {
                 for (i = 0; i < buffer1->used; ++i)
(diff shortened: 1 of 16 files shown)


libutils/buffer.c
```diff
@@ -333,7 +333,8 @@ void BufferAppendF(Buffer *buffer, const char *format, ...)
     va_copy(aq, ap);
 
     int printed = vsnprintf(buffer->buffer + buffer->used, buffer->capacity - buffer->used, format, aq);
-    if (printed >= (buffer->capacity - buffer->used))
+    assert(printed >= 0);
+    if ((unsigned int) printed >= (buffer->capacity - buffer->used))
```
vpodzime, line 337: I'd prefer changing `buffer->capacity/used` to `size_t` and then work with a `size_t printed` variable here.

libutils/hash.c
```diff
@@ -257,15 +257,15 @@ Hash *HashNewFromKey(const RSA *rsa, HashMethod method)
     }
 
     unsigned char buffer[buf_len];
-    int actlen;
+    size_t actlen;
 
     actlen = BN_bn2bin(n, buffer);
-    CF_ASSERT(actlen <= buf_len, "Buffer overflow n, %d > %zu!",
+    CF_ASSERT(actlen <= buf_len, "Buffer overflow n, %ld > %zu!",
```
vpodzime, line 263: The format for `size_t` is `%zd` IIRC.

vpodzime (commented):
Looks good to me otherwise.

libutils/hash.c
olehermanse: `size_t` is unsigned, so should be `%zu`;

https://github.com/cfengine/core/blob/master/CONTRIBUTING.md#string-formatting

libutils/buffer.c
olehermanse: Just be absolutely sure that `buffer->capacity/used` is never below 0, then(!). For example in while/for loops.

larsewi: @olehermanse  `buffer->capacity/used` is already `unsigned`. So that should not be a problem.

libutils/hash.c
vpodzime: Correct, good catch. :+1:

libutils/buffer.c
```diff
@@ -32,7 +32,7 @@
 #endif
```
olehermanse, line 32: @larsewi This commit doesn't have any user-facing changes, so there should not be a changelog entry.

https://github.com/cfengine/core/blob/master/CONTRIBUTING.md#changelog

olehermanse (changes requested):
Looks good overall, just change the commit message to have

```
Changelog: None
```

vpodzime (approved):
ACK

olehermanse (approved):

libutils/buffer.c
```diff
@@ -374,7 +375,11 @@ int BufferPrintf(Buffer *buffer, const char *format, ...)
      * everything is easy, however if we are shared then we need a different strategy.
      */
     int printed = vsnprintf(buffer->buffer, buffer->capacity, format, aq);
-    if (printed >= buffer->capacity)
+    if (printed < 0) 
+    {
+        // vsnprintf failed!
```
craigcomstock, line 380: @larsewi @olehermanse @vpodzime I wonder about these comments? Should it be a TODO to do something about the case where vsnprintf fails?

larsewi: @craigcomstock in this case `BufferPrintf` returns the negative value of `vsnprintf` when it failes. Thus the caller of `BufferPrintf` must handle it. I don't think it makes sence to handle such errors in a utillity function.

PR #114, opened 2021-01-29
Removed unnecessary assert comparison

larsewi:
Variable of unsigned type was checked to be greater that zero.

Changelog: None
Ticket: None


vpodzime (approved):
ACK

olehermanse (approved):

PR #115, opened 2021-02-12
Always use xmalloc() in threaded_stack.c

vpodzime:
To make sure the process is killed in case of no memory left
instead of trying to work with NULL returned from malloc().


olehermanse (approved):

vpodzime:
> You’re not [authorized](https://docs.github.com/articles/about-protected-branches/) to merge this pull request.

I cannot merge this for some reason.

PR #116, opened 2021-02-25
Add a new NDEBUG_UNUSED macro for variables only used in debug builds

vpodzime:
Sometimes a variables is only used in an assertion like
'assert(ret > 0)` which results in a compilation warning/error
when compiling without '--enable-debug' because in such cases
all 'assert()' are removed.

### DIFF
--- a/libutils/compiler.h
+++ b/libutils/compiler.h
@@ -70,6 +70,16 @@
 
 
 /**
+ * For variables only used in debug builds, in particular only in assert()
+ * calls, use NDEBUG_UNUSED.
+ */
+#ifdef NDEBUG
+#define NDEBUG_UNUSED __attribute__((unused))
+#else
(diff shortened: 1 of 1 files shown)


olehermanse (approved):
Makes sense to me. I guess you could use the UNUSED macro, but this is more specific.

It does stand out to me that you are using a different method than the other macro (void cast vs. unused attribute) - thoughts on whether this attribute works _everywhere_?

vpodzime:
> thoughts on whether this attribute works _everywhere_?

Let's see: [Build Status](https://ci.cfengine.com/job/pr-pipeline/6146/)

vpodzime:
Also passed _static-checks_ [Build Status](https://ci.cfengine.com/job/static-checks/623/) (incl. _clang_ compilation) so I think this is safe enough.

PR #117, opened 2021-03-09
Return NULL from FileWriter() and CsvWriterOpen() if they get NULL

vpodzime:
This makes use of the two "constructors" easier because the code
can simply look like this:

  FILE *file = fopen(...);
  Writer *writer = FileWriter(file);
  CsvWriter *csv_writer = CsvWriterOpen(writer);

  if (csv_writer == NULL)
  {
      /* handle error */
  }

instead of checking each return value for != NULL. It usually
doesn't matter much which phase exactly failed.

Ticket: CFE-3554
Changelog: None

### DIFF
--- a/libutils/csv_writer.c
+++ b/libutils/csv_writer.c
@@ -52,6 +52,10 @@ CsvWriter *CsvWriterOpenSpecifyTerminate(Writer *w, bool terminate_last_line)
 
 CsvWriter *CsvWriterOpen(Writer *w)
 {
+    if (w == NULL)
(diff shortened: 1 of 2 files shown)


larsewi (approved):
ACK

vpodzime:
Turns out `FileWriter(NULL);` is actually abused in the unit tests: https://github.com/cfengine/libntech/blob/master/tests/unit/file_writer_test.c#L36
Changing that is out of scope for me now.

PR #120, opened 2021-03-17
Add JsonWalk function

vpodzime:
A function to go through all the elements in a given JSON and do
arbitrary operations on them. Useful for validation or any
functional-style computation on a JSON.

Ticket: ENT-6789
Changelog: None

### DIFF
--- a/libutils/json.c
+++ b/libutils/json.c
@@ -2749,6 +2749,61 @@ JsonParseError JsonParseFile(
     return JsonParseAnyFile(path, size_max, json_out, false);
 }
 
+bool JsonWalk(JsonElement *element,
+              JsonElementVisitor object_visitor,
+              JsonElementVisitor array_visitor,
+              JsonElementVisitor primitive_visitor,
+              void *data)
+{
+    assert(element != NULL);
+
+    if (element->type == JSON_ELEMENT_TYPE_PRIMITIVE)
+    {
+        if (primitive_visitor != NULL)
+        {
+            return primitive_visitor(element, data);
+        }
+        else
+        {
+            return true;
+        }
+    }
+
+    assert(element->type == JSON_ELEMENT_TYPE_CONTAINER);
+    if (element->container.type == JSON_CONTAINER_TYPE_ARRAY)
+    {
+        if (array_visitor != NULL)
+        {
+            bool keep_going = array_visitor(element, data);
+            if (!keep_going)
+            {
+                return false;
+            }
+        }
+    }
+    else
+    {
+        assert(element->container.type == JSON_CONTAINER_TYPE_OBJECT);
+        if (object_visitor != NULL)
+        {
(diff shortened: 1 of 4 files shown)


larsewi (approved):
These Boots Are Made For Walkin'

tests/unit/json_test.c
```diff
@@ -847,6 +847,78 @@ static void test_iterator_current(void)
...
+    return true;
+}
+
+static bool VisitArray(JsonElement *array, void *data)
+{
+    assert_int_equal(JsonGetElementType(array), JSON_ELEMENT_TYPE_CONTAINER);
+    assert_int_equal(JsonGetContainerType(array), JSON_CONTAINER_TYPE_ARRAY);
+
+    Writer *writer = data;
+    WriterWriteF(writer, "[%zd]", JsonLength(array));
```
olehermanse, line 866:
```suggestion
    WriterWriteF(writer, "[%zd],", JsonLength(array));
```

Adding a comma here would make the trace more understandable, at least for me.

(Remember to also update expected trace below if you change this).

libutils/json.c
```diff
@@ -2749,6 +2749,62 @@ JsonParseError JsonParseFile(
...
+        {
+            return primitive_visitor(element, data);
+        }
+        else
+        {
+            return true;
+        }
+    }
+    else
+    {
```
olehermanse, line 2772: Since the if branches always return, these `else` blocks are unnecessary. If you dropped them all the code below would be de-indented, and you would have a `return` at the end which is not inside an extra scope. I'd prefer that, but no big deal.

tests/unit/json_test.c
```diff
@@ -847,6 +847,78 @@ static void test_iterator_current(void)
...
+    return true;
+}
+
+static bool VisitArrayAbortOnEmpty(JsonElement *array, void *data)
+{
+    assert_int_equal(JsonGetElementType(array), JSON_ELEMENT_TYPE_CONTAINER);
+    assert_int_equal(JsonGetContainerType(array), JSON_CONTAINER_TYPE_ARRAY);
+
+    Writer *writer = data;
+    WriterWriteF(writer, "[%zd]", JsonLength(array));
```
olehermanse, line 893:
```suggestion
    WriterWriteF(writer, "[%zd],", JsonLength(array));
```

Comma here would make the trace more readable, IMO.

olehermanse (approved):
Very nice overall!

tests/unit/json_test.c
vpodzime: I like how commas only separate values. `[2]value1,value2,` says those two values are in an array. But in general this is not at all important. A proper trace would have to be smarter.

libutils/json.c
vpodzime: Good point!

olehermanse (approved):

PR #126, opened 2021-03-31
CFE-3416: Fixed implicit-fallthrough compiler warnings on multiple files

KotzaBoss:
In an attempt to silence deliberate fall throughs
while being portable, an one line comment `/* fall through */`
was inserted at the end of the fall through-ing case.

### DIFF
--- a/libutils/json.c
+++ b/libutils/json.c
@@ -2019,7 +2019,9 @@ static JsonParseError JsonParseAsString(
             switch (**data)
             {
             case '\\':
+                break;
             case '"':
+                break;
             case '/':
                 break;
 
@@ -2052,7 +2054,7 @@ static JsonParseError JsonParseAsString(
                 WriterWriteChar(writer, '\\');
                 break;
             }
-            /* Deliberate fall-through */
+            /* fall through */
         default:
             WriterWriteChar(writer, **data);
             break;
--- a/libutils/mustache.c
+++ b/libutils/mustache.c
@@ -252,6 +252,7 @@ static Mustache NextTag(const char *input,
         break;
     case '{':
         extra_end = "}";
+        // fall through
     case '&':
         ret.type = TAG_TYPE_VAR_UNESCAPED;
         ret.content++;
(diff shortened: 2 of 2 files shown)


libutils/json.c
```diff
@@ -2019,7 +2019,9 @@ static JsonParseError JsonParseAsString(
             switch (**data)
             {
             case '\\':
+                // fall through
             case '"':
+                // fall through
```
olehermanse, line 2024: Since you are only falling through to `break` here, you could consider just adding `break;` instead.

libutils/mustache.c
```diff
@@ -252,6 +252,7 @@ static Mustache NextTag(const char *input,
         break;
     case '{':
         extra_end = "}";
+        break;
```
olehermanse, line 255: So, here you've changed behavior, you changed something which was a fall through, to a break. Why did you make this change? There is no explaining comment, or explanation in the commit message.

libutils/json.c
KotzaBoss: indeed i was on autopilot when i wrote this and its silly, will fix

libutils/mustache.c
KotzaBoss: This is a mis-reading of the code on my part it seems. Seeing the rest of the fall throughs had a comment explaining that they are intentional i, apparently wrongly, assumed it was _not_ intentional and the break was missing.
Looking back at it i could have inferred that it was (intentional) because there is a pattern of breaking after setting up the `ret`. It should be changed to `/* fall through */`

olehermanse: Makes sense. Next time, double check that you understand what the code is doing when making a change like this, and there should be an explanation as well.

vpodzime (dismissed):
Looks good to me. @KotzaBoss make sure to resolve the CLA issues.

KotzaBoss:
All set!
Do you think it would be a good idea to mention the deliberate fall through comment in [CONTRIBUTING](https://github.com/cfengine/core/blob/master/CONTRIBUTING.md#coding-style)?

I saw that when it was deliberate people put _a_ comment but not one to save future silencing of warnings.

vpodzime:
>  Do you think it would be a good idea to mention the deliberate fall through comment in [CONTRIBUTING](https://github.com/cfengine/core/blob/master/CONTRIBUTING.md#coding-style)?

Yup, that's a good point! Feel free to add a commit here or open a new PR with the change.

olehermanse (changes requested):
Both PR description and commit message have this:

```
Ticket: Title
```

Replace `Title` there with the ticket number from JIRA (CFE-3416). Also I've updated the PR title for you to start with the ticket number.

vpodzime (approved):
ACK

olehermanse (changes requested):
Almost there, you fixed the commit message, but not the PR description. (The first comment).

olehermanse (approved):
Great, thanks!

PR #127, opened 2021-04-06
Added absolute value macro

larsewi:
Ticket: None
Changelog: None

### DIFF
--- a/libutils/misc_lib.h
+++ b/libutils/misc_lib.h
@@ -73,6 +73,8 @@
 
 #define ISPOW2(n)     (  (n)>0  &&  ((((n) & ((n)-1)) == 0))  )
 
+#define ABS(n)        (  (n<0)  ?  (-n)  :  (n)  )
+
 
 /*
   In contrast to the standard C modulus operator (%), this gives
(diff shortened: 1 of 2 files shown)


libutils/misc_lib.h
```diff
@@ -73,6 +73,8 @@
 
 #define ISPOW2(n)     (  (n)>0  &&  ((((n) & ((n)-1)) == 0))  )
 
+#define ABS(n)        (  (n<0)  ?  (-n)  :  (n)  )
```
olehermanse, line 76: Note that `ABS(somefunc())` will cause `somefunc()` to be called 3 times - which can be significant if it's slow. Not much to do about it in C, since we don't know the type, but useful to know / pay attention to. (in C++ we could create a temp variable with `auto` type, I think)

olehermanse (approved):
🚀

vpodzime (approved):
ACK

libutils/misc_lib.h
larsewi: Intresting @olehermanse . Never thought about that, but makes sense. Thanks for the heads up 👍

PR #129, opened 2021-04-18
Added autoconf message for whether debug option is present

olehermanse:
This option was useful to confirm that libntech is built
with debug options. I though I might leave it in.


vpodzime (approved):
ACK

PR #130, opened 2021-04-20
WIP: Add a process manager "class"

vpodzime:
In many places there's a need to keep track of sub-processes
spawned by the main process. Such processes need to be looked up
by various parameters, interacted with, and, eventually,
terminated.

Having a common, reliable and fast implementation with high test
coverage that holds the various pieces of information about
sub-processes together is beneficial.

Ticket: CFE-3572
Changelog: None

Steps:
- [x] initialization
- [x] adding subprocesses
- [x] getter functions
- [x] pop functions
- [x] termination functions
- [ ] fix issues on exotics
- [ ] add `void *data` to the `Subprocess` struct

### DIFF
--- a/libutils/array_map.c
+++ b/libutils/array_map.c
@@ -86,6 +86,26 @@ bool ArrayMapRemove(ArrayMap *map, const void *key)
     return false;
 }
 
+bool ArrayMapRemoveSoft(ArrayMap *map, const void *key)
+{
+    assert(map != NULL);
+
+    for (int i = 0; i < map->size; ++i)
+    {
+        if (map->equal_fn(map->values[i].key, key))
+        {
+            map->destroy_key_fn(map->values[i].key);
+
+            memmove(map->values + i, map->values + i + 1,
+                    sizeof(MapKeyValue) * (map->size - i - 1));
+
+            map->size--;
+            return true;
+        }
+    }
+    return false;
+}
+
 MapKeyValue *ArrayMapGet(const ArrayMap *map, const void *key)
 {
     for (int i = 0; i < map->size; ++i)
--- a/libutils/hash_map.c
+++ b/libutils/hash_map.c
@@ -172,6 +172,39 @@ bool HashMapRemove(HashMap *map, const void *key)
     return false;
 }
 
+bool HashMapRemoveSoft(HashMap *map, const void *key)
+{
+    assert(map != NULL);
+
+    unsigned bucket = HashMapGetBucket(map, key);
+
+    /*
+     * prev points to a previous "next" pointer to rewrite it in case value need
+     * to be deleted
+     */
+
+    for (BucketListItem **prev = &map->buckets[bucket];
+         *prev != NULL;
+         prev = &((*prev)->next))
+    {
+        BucketListItem *cur = *prev;
+        if (map->equal_fn(cur->value.key, key))
+        {
+            map->destroy_key_fn(cur->value.key);
+            *prev = cur->next;
+            free(cur);
+            map->load--;
+            if ((map->load < map->min_threshold) && (map->size > map->init_size))
+            {
+                HashMapResize(map, map->size >> 1);
+            }
+            return true;
+        }
+    }
+
+    return false;
+}
+
 MapKeyValue *HashMapGet(const HashMap *map, const void *key)
 {
     unsigned bucket = HashMapGetBucket(map, key);
--- a/libutils/proc_manager.c
+++ b/libutils/proc_manager.c
@@ -1,0 +1,618 @@
+/*
+  Copyright 2021 Northern.tech AS
+
+  This file is part of CFEngine 3 - written and maintained by Northern.tech AS.
+
+  This program is free software; you can redistribute it and/or modify it
+  under the terms of the GNU General Public License as published by the
+  Free Software Foundation; version 3.
+
+  This program is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with this program; if not, write to the Free Software
+  Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307, USA
+
+  To the extent this program is licensed as part of the Enterprise
+  versions of CFEngine, the applicable Commercial Open Source License
+  (COSL) may apply to this file if you as a licensee so wish it. See
+  included file COSL.txt.
+*/
+
+#include <platform.h>
+
+#include <map.h>                /* Map* */
+#include <alloc.h>              /* xmalloc, xasprintf */
+#include <string_lib.h>         /* StringHash_untyped, StringEqual_untyped */
+#include <logging.h>
+#include <proc_manager.h>
+
+#ifndef __MINGW32__
+#include <errno.h>
+#include <signal.h>             /* kill, SIGKILL */
+#endif
+
+#ifdef __MINGW32__
+#define fileno _fileno
+#endif
+
+#define SIGKILL_TERMINATION_TIMEOUT 5 /* seconds */
+
+struct ProcManager_ {
+    Map *procs_by_id;  /** primary storage for the subprocesses info (owns the data)    */
+    Map *procs_by_fd;  /** a different view on the data to speed up lookup by FD/stream */
+    Map *procs_by_pid; /** a different view on the data to speed up lookup by PID       */
+};
+
+static unsigned int HashFD_untyped(const void *fd, ARG_UNUSED unsigned int seed)
+{
+    assert(fd != NULL);
+    return *((unsigned int*) fd);
+}
+
+static bool FDEqual_untyped(const void *fd1, const void *fd2)
+{
+    assert((fd1 != NULL) && (fd2 != NULL));
+    return (*((int*) fd1) == *((int*) fd2));
+}
+
+static unsigned int HashPID_untyped(const void *pid, ARG_UNUSED unsigned int seed)
+{
+    assert(pid != NULL);
+    return *((unsigned int*) pid);
+}
+
+static bool PIDEqual_untyped(const void *pid1, const void *pid2)
+{
+    assert((pid1 != NULL) && (pid2 != NULL));
+    return (*((pid_t*) pid1) == *((pid_t*) pid2));
+}
+
+static void NoopDestroy(ARG_UNUSED void *p)
+{
+    /* not destroying anything here */
+}
+
+ProcManager *ProcManagerNew()
+{
+    ProcManager *manager = xmalloc(sizeof(ProcManager));
+
+    /* keys are just pointers into data, this map actually holds the data */
+    manager->procs_by_id = MapNew(StringHash_untyped, StringEqual_untyped,
+                                  NoopDestroy, (MapDestroyDataFn) SubprocessDestroy);
+
+    /* keys are allocated, data belongs to the procs_by_id map */
+    manager->procs_by_fd = MapNew(HashFD_untyped, FDEqual_untyped,
+                                  free, NoopDestroy);
+
+    /* keys are just pointers into data, data belongs to the procs_by_id map */
+    manager->procs_by_pid = MapNew(HashPID_untyped, PIDEqual_untyped,
+                                   NoopDestroy, NoopDestroy);
+
+    return manager;
+}
+
+void ProcManagerDestroy(ProcManager *manager)
+{
+    if (manager != NULL)
+    {
+        /* procs_by_id owns the data */
+        MapSoftDestroy(manager->procs_by_fd);
+        MapSoftDestroy(manager->procs_by_pid);
+        MapDestroy(manager->procs_by_id);
+        free(manager);
+    }
+}
+
+static inline Subprocess *SubprocessNew(char *id, char *cmd, char *description,
+                                        pid_t pid, FILE *input, FILE *output, char lookup_io)
+{
+    Subprocess *ret  = xmalloc(sizeof(Subprocess));
+    ret->id          = id;
+    ret->cmd         = cmd;
+    ret->description = description;
+    ret->pid         = pid;
+    ret->input       = input;
+    ret->output      = output;
+    ret->lookup_io   = lookup_io;
+
+    return ret;
+}
+
+void SubprocessDestroy(Subprocess *proc)
+{
+    if (proc != NULL)
+    {
+        free(proc->id);
+        free(proc->cmd);
+        free(proc->description);
+        free(proc);
+    }
+}
+
+static inline int *DupInt(int val)
+{
+    int *ret = xmalloc(sizeof(int));
+    *ret = val;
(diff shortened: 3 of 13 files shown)


libutils/proc_manager.h
```diff
@@ -0,0 +1,123 @@
...
+
+/**
+ * Get a new ProcManager.
+ */
+ProcManager *ProcManagerNew();
+
+/**
+ * Destroy a ProcManager.
+ */
+void *ProcManagerDestroy();
```
KotzaBoss, line 70: Why is the return value of destroy `void*`?

libutils/proc_manager.h
```diff
@@ -0,0 +1,123 @@
...
+bool ProcManagerAddProcess(ProcManager *manager,
+                           char *id, char *cmd, char *description,
+                           pid_t pid, int in_fd, int out_fd, char lookup_fd);
+
+/**
+ * Getter functions for #ProcManager.
+ */
+Subprocess *ProcManagerGetProcessByPID(ProcManager *manager, pid_t pid);
+Subprocess *ProcManagerGetProcessById(ProcManager *manager, const char *id);
+Subprocess *ProcManagerGetProcessByFD(ProcManager *manager, int fd);
```
KotzaBoss, line 95: could multiple processes share the same fd?
This question is shared for the other functions like pop and terminate too.

libutils/proc_manager.h
```diff
@@ -0,0 +1,123 @@
...
+ * Struct to hold information about a subprocess.
+ */
+struct _Subprocess {
+    char *id;             /** unique ID of the subprocess                  [always not %NULL] */
+    char *cmd;            /** command used to create the subprocess        [can be %NULL] */
+    char *description;    /** human-friendly description of the subprocess [can be %NULL] */
+    pid_t pid;            /** PID of the subprocess                        [always >= 0] */
+    int in_fd;            /** FD of the input to the subprocess            [-1 if N/A] */
+    int out_fd;           /** FD of the output from the subprocess         [-1 if N/A] */
+    char lookup_fd;       /** which FD to use for lookup by FD             ['i', 'o' or '\0'] */
```
KotzaBoss, line 40: perhaps a simple enum to give a bit more context to the options?
```c
enum LookupByFD {
    IN,
    OUT,
    NONE
};

File: libutils/proc_manager.h (hunk 1 above)
Comment (vpodzime) [reply libutils/proc_manager.h:70]: Good catch, should be `void` of course.

File: libutils/proc_manager.h (hunk 2 above)
Comment (vpodzime) [reply libutils/proc_manager.h:95]: No, FDs in one process are unique, otherwise `write(fd,...)` wouldn't know where to write to.

File: libutils/proc_manager.h (hunk 3 above)
Comment (vpodzime) [reply libutils/proc_manager.h:40]: That was my original idea too, but `enum` is an `int` in C, i.e. at least 4 bytes big. `char` is just one byte.

File: libutils/proc_manager.h [hunk 4]
```diff
@@ -0,0 +1,123 @@
...
+Subprocess *ProcManagerGetProcessByFD(ProcManager *manager, int fd);
+
+/**
+ * Functions to get and remove subprocesses from a #ProcManager.
+ *
+ * @note Destroy the returned #Subprocess with SubprocessDestroy().
+ */
+Subprocess *ProcManagerPopProcessByPID(ProcManager *manager, pid_t pid);
+Subprocess *ProcManagerPopProcessById(ProcManager *manager, const char *id);
+Subprocess *ProcManagerPopProcessByFD(ProcManager *manager, int fd);
```
larsewi, line 104: Is not the term "pop" related to stacks? Could perhaps "remove" be better phrasing?

vpodzime: I think *pop* is quite generic for the case where you *get* something while *removing* it at the same time.

libutils/proc_manager.h
olehermanse: If we were using C++ (C++ 11 or later) we could:

```C++
enum LookupByFD: uint8_t
{
    IN,
    OUT,
    NONE
}
```

Without it, the "workaround" is:

```C++
enum _LookupByFD
{
    IN,
    OUT,
    NONE
}

typedef LookupByFD uint8_t;
```

But I agree with your choice, this is not much "nicer".

vpodzime:
I replaced the file descriptors by `FILE` streams. Our code works with them and while it's possible to go from a `FILE` stream to an FD (`fileno()`), the other direction is impossible. Lookup functions "by FD" still make sense to me, though.

vpodzime:
Actually, we will need the termination functionality for custom promise modules.

libutils/array_map.c
```diff
@@ -86,6 +86,26 @@ bool ArrayMapRemove(ArrayMap *map, const void *key)
     return false;
 }
 
+bool ArrayMapRemoveSoft(ArrayMap *map, const void *key)
+{
+    assert(map != NULL);
+
+    for (int i = 0; i < map->size; ++i)
+    {
+        if (map->equal_fn(map->values[i].key, key))
```
KotzaBoss, line 95: I looked up the declarations for the `ArrayMap` and i have a series of interlinked questions.
- Why would each `ArrayMap` "instance" need to have a pointer to the equal, destroy etc functions?
- How is it expected to be used?
- Doesnt the `ArrayMap` have _a_ specific implementation?

I saw the use of `ArrayMapNew` [here](https://github.com/cfengine/libntech/blob/828aa81863f96f243ca3b8a824e1e154f7878df1/libutils/map.c#L107) and [here](https://github.com/cfengine/libntech/blob/828aa81863f96f243ca3b8a824e1e154f7878df1/tests/unit/map_test.c#L531) but im still not sure i _exactly_ get the idea behind giving the map these three functions.

vpodzime: Well, the map needs to be able to compare the keys, destroy the keys and destroy data. It's a generic key-value storage so it stores `void` pointers and needs to know how to compare them and how to destroy the data they point to.

vpodzime:
@olehermanse please review, if you have a chance. The only missing bit here, AFAICT, are unit tests for the termination functions.

libutils/hash_map.c
```diff
@@ -172,6 +172,40 @@ bool HashMapRemove(HashMap *map, const void *key)
...
+    {
+        BucketListItem *cur = *prev;
+        if (map->equal_fn(cur->value.key, key))
+        {
+            map->destroy_key_fn(cur->value.key);
+            *prev = cur->next;
+            free(cur);
+            map->load--;
+            if ((map->load < map->min_threshold) && (map->size > map->init_size))
+
```
larsewi, line 198: white space

vpodzime: ? I don't see any white space problem in the above. Moreover, it's just a code copied from the `MapRemove()` function.

larsewi: I just though an additional newline between the condition and the if-block looked kind of weird. But I guess it's alright. 👍

vpodzime: Yeah, like I said copy-pasted code. But I fixed it here at least. Good catch!

vpodzime:
@cf-bottom jenkins, please

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/6479/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/6479/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-6479/

libutils/proc_manager.c
```diff
@@ -0,0 +1,614 @@
...
+{
+    assert(manager != NULL);
+    assert(proc != NULL);
+
+    MapRemoveSoft(manager->procs_by_pid, &(proc->pid));
+
+    if (lookup_fd != -1)
+    {
+        MapRemoveSoft(manager->procs_by_fd, &lookup_fd);
+    }
```
olehermanse, line 318: Wouldn't `MapRemove()` be the same, since they are using `NoopDestroy()`?

libutils/proc_manager.c
```diff
@@ -0,0 +1,614 @@
...
+    assert(manager != NULL);
+
+    int fd = fileno(stream);
+    Subprocess *proc = MapGet(manager->procs_by_fd, &fd);
+    SoftRemoveProcFromMaps(manager, proc);
+    return proc;
+}
+
+
+static inline void RemoveProcFromMaps(ProcManager *manager, Subprocess *proc, int lookup_fd)
```
olehermanse, line 308: It seems a little backwards / error prone to me that `lookup_fd` is a parameter to this function (separate to the subprocess). `SoftRemoveProcFromMaps()` does not have this parameter.

olehermanse (dismissed):
Looks good in general, just some smaller comments.

libutils/array_map.c
```diff
@@ -86,6 +86,26 @@ bool ArrayMapRemove(ArrayMap *map, const void *key)
...
+        {
+            map->destroy_key_fn(map->values[i].key);
+
+            memmove(map->values + i, map->values + i + 1,
+                    sizeof(MapKeyValue) * (map->size - i - 1));
```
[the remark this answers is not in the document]
vpodzime: It's just a copy-pasted code from `ArrayMapRemove()`.

libutils/proc_manager.c
```diff
@@ -0,0 +1,614 @@
...
+Subprocess *ProcManagerPopProcessById(ProcManager *manager, const char *id)
+{
+    assert(manager != NULL);
+
+    Subprocess *proc = MapGet(manager->procs_by_id, id);
+    SoftRemoveProcFromMaps(manager, proc);
+    return proc;
+}
+
+Subprocess *ProcManagerPopProcessByFD(ProcManager *manager, int fd)
```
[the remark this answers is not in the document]
vpodzime: Yes, it's the same as `ProcManagerGetProcessByFD()` which is tested, combined with removal of the result from the internal maps which is tested by the test for `ProcManagerPopProcessByPID()`. That's good enough for me.

libutils/proc_manager.c
```diff
@@ -0,0 +1,614 @@
...
+Subprocess *ProcManagerPopProcessByPID(ProcManager *manager, pid_t pid)
+{
+    assert(manager != NULL);
+
+    Subprocess *proc = MapGet(manager->procs_by_pid, &pid);
+    SoftRemoveProcFromMaps(manager, proc);
+    return proc;
+}
+
+Subprocess *ProcManagerPopProcessById(ProcManager *manager, const char *id)
```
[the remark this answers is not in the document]
vpodzime: Same here.

libutils/array_map.c
```diff
@@ -86,6 +86,26 @@ bool ArrayMapRemove(ArrayMap *map, const void *key)
...
+
+            map->size--;
+            return true;
+        }
+    }
+    return false;
```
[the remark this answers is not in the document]
vpodzime: > Your proc manager tests always use ArrayMap, never HashMap

Yeah, there are only few (fake) processes. I'll add a test with more.

libutils/proc_manager.c
```diff
@@ -0,0 +1,614 @@
...
+static unsigned int HashFD_untyped(const void *fd, ARG_UNUSED unsigned int seed)
+{
+    assert(fd != NULL);
+    return *((unsigned int*) fd);
+}
+
+static bool FDEqual_untyped(const void *fd1, const void *fd2)
+{
+    assert((fd1 != NULL) && (fd2 != NULL));
+    return (*((int*) fd1) == (*(int*) fd2));
```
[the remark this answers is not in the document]
vpodzime: I think it's actually clearer the way it is on the left side, saying "dereference this pointer to int" instead of "dereference this pointer type and type-cast the variable".

libutils/proc_manager.c
vpodzime: Yes, it would. But no need to do those unnecessary function calls. :wink:

libutils/proc_manager.c
vpodzime: Because `SoftRemoveProcFromMaps()` is not used in the place where the `input/output` streams are closed already and thus unusable for `fileno()`. I'll add a comment.

libutils/array_map.c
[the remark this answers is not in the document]
vpodzime: Actually, we don't have any unit tests doing `ArrayMapRemove()`. I created [CFE-3643[(https://tracker.mender.io/browse/CFE-3643) instead of trying to save the world in this PR.

libutils/proc_manager.c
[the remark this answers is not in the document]
olehermanse: So it should be changed on the right side.

vpodzime:
@cf-bottom jenkins with exotics, please

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/6482/)

(with exotics)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/6482/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-6482/

vpodzime:
There are failures on exotics. And we also need to be able to associate some custom data with the sub-processes managed.

olehermanse (approved):
Looks good to me, thanks for working on this! I see there are some remaining TODOs, but feel free to address those in separate PRs.

olehermanse (changes requested):
> There are failures on exotics. And we also need to be able to associate some custom data with the sub-processes managed.

Need to be addressed one way or another, before merging.

libutils/proc_manager.h
```diff
@@ -0,0 +1,139 @@
...
+ * Struct to hold information about a subprocess.
+ */
+struct Subprocess_ {
+    char *id;             /** unique ID of the subprocess                    [always not %NULL] */
+    char *cmd;            /** command used to create the subprocess          [can be %NULL]     */
+    char *description;    /** human-friendly description of the subprocess   [can be %NULL]     */
+    pid_t pid;            /** PID of the subprocess                          [always >= 0]      */
+    FILE *input;          /** stream of the input to the subprocess          [can be %NULL]     */
+    FILE *output;         /** stream of the output from the subprocess       [can be %NULL]     */
+    char lookup_io;       /** which stream/FD to use for lookup by stream/FD ['i', 'o' or '\0'] */
```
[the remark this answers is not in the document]
vpodzime: This depends on the direction of the communication with the particular subprocess.

libutils/proc_manager.h
```diff
@@ -0,0 +1,139 @@
...
+ * @param description human-friendly description of the subprocess   [can be %NULL]
+ * @param pid         PID of the subprocess                          [required to be >=0]
+ * @param input       stream of the input to the subprocess          [can be %NULL]
+ * @param output      stream of the output from the subprocess       [-1 if N/A]
+ * @param lookup_io   which stream/FD to use for lookup by stream/FD ['i', 'o' or '\0']
+ *
+ * @return whether the subprocess was successfully added or not (e.g. in case of
+ *         the process with the same id, PID or lookup stream/FD added before)
+ *
+ * @note Ownership of #id, #cmd and #description is transferred to the #manager.
```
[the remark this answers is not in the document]
vpodzime: Only in case of termination.

PR #132, opened 2021-04-29
Fix ambiguous definition of GetDefaultDir_helper()

vpodzime:
It should use an array in both the declaration and the
definition.


larsewi (approved):
ACK

PR #135, opened 2021-04-29
Added StringDecimalToLong utility function

KotzaBoss:
Its primary use is the int policy function which
may accept floating point numbers as strings.

Ticket: None
Changelog: None

### DIFF
--- a/libutils/string_lib.c
+++ b/libutils/string_lib.c
@@ -32,7 +32,7 @@
 #include <cleanup.h>
 #include <definitions.h> // CF_BUFSIZE
 #include <condition_macros.h> // nt_static_assert()
-
+#include <printsize.h>
 
 char *StringVFormat(const char *fmt, va_list ap)
 {
@@ -529,8 +529,43 @@ int StringToLong(const char *str, long *value_out)
 }
 
 /**
- * @brief Log StringToLong & StringToUlong conversion error with an identifier
- *        for debugging.
+ * @brief Convert a string representing a real number to int, discarding the decimal part.
+ *
+ * Attempts to find the first '.' character in order to copy the integer part to a
+ * temporary array before passing it to StringToLong. Should the input string be
+ * an entire integer by itself no copying occurs. The integer part must NOT exceed 12 digits.
+ *
+ * @param[in] str String matching the regex \d+(\.\d+)?, cannot be NULL
+ * @param[out] value_out Where to store result on success, cannot be NULL
+ * @return 0 on success, -84 if the integer part is too long or other error code otherwise (see StringToLong source code)
+ */
+int StringDecimalToLong(const char *str, long *value_out)
+{
+    assert(str != NULL);
+    assert(value_out != NULL);
+
+    size_t len_of_int_part = strcspn(str, ".");
+
+    if (len_of_int_part > PRINTSIZE(INT_MAX))
+    {
+        return -84;  // Integer part too large to allocate buffer
+    }
+
+    if (len_of_int_part == 0 || str[len_of_int_part] == '\0')
+    {
+        return StringToLong(str, value_out);
+    }
+
+    char int_part[len_of_int_part + 1];
+    strncpy(int_part, str, len_of_int_part);
+    int_part[len_of_int_part] = '\0';
+
+    return StringToLong(int_part, value_out);
+}
+
+/**
+ * @brief Log StringToLong & StringToUlong & StringDecimalToLong conversion error
+ *        with an identifier for debugging.
  */
 void LogStringToLongError(const char *str_attempted, const char *id, int error_code)
 {
@@ -549,6 +584,9 @@ void LogStringToLongError(const char *str_attempted, const char *id, int error_c
             break;
         case -83:
             error_str = "Not terminated";
+            break;
+        case -84:
+            error_str = "Integer part too large";
             break;
     }
     Log(LOG_LEVEL_ERR, "Conversion error (%d - %s) on '%s' (%s)", error_code, error_str, str_attempted, id);
--- a/tests/unit/string_lib_test.c
+++ b/tests/unit/string_lib_test.c
@@ -496,6 +496,76 @@ static void test_string_to_long_compatibility(void)
     // Old function (StringToLongUnsafe) does not allow trailing whitespace
 }
 
+#define ASSERT_INT_EQUAL_STRING_DECIMAL_TO_LONG(str, value_out, expected) \
+    assert_int_equal(StringDecimalToLong(str, &value_out), 0); \
+    assert_int_equal(value_out, expected)
+
+static void test_string_decimal_to_long(void)
+{
+    long value_out;
+
+    // Basic usage:
+    ASSERT_INT_EQUAL_STRING_DECIMAL_TO_LONG("0.0", value_out, 0);
+    ASSERT_INT_EQUAL_STRING_DECIMAL_TO_LONG("-0.0", value_out, 0);
+    ASSERT_INT_EQUAL_STRING_DECIMAL_TO_LONG("+0.0", value_out, 0);
+    ASSERT_INT_EQUAL_STRING_DECIMAL_TO_LONG("123.0", value_out, 123);
+    ASSERT_INT_EQUAL_STRING_DECIMAL_TO_LONG("+123.0", value_out, 123);
+    ASSERT_INT_EQUAL_STRING_DECIMAL_TO_LONG("-123.0", value_out, (long) -123);
+
+    // Leading spaces:
+    ASSERT_INT_EQUAL_STRING_DECIMAL_TO_LONG(" 0.0", value_out, 0);
+    ASSERT_INT_EQUAL_STRING_DECIMAL_TO_LONG(" 123.0", value_out, 123);
+    ASSERT_INT_EQUAL_STRING_DECIMAL_TO_LONG(" -123.0", value_out, (long) -123);
+    ASSERT_INT_EQUAL_STRING_DECIMAL_TO_LONG("           0.0", value_out, 0);  // 12 is max accepted int length
+    ASSERT_INT_EQUAL_STRING_DECIMAL_TO_LONG("      123.0", value_out, 123);
+    ASSERT_INT_EQUAL_STRING_DECIMAL_TO_LONG("      -123.0", value_out, (long) -123);
+
+    // Trailing spaces:
+    ASSERT_INT_EQUAL_STRING_DECIMAL_TO_LONG("9.0 ", value_out, 9);
+    ASSERT_INT_EQUAL_STRING_DECIMAL_TO_LONG("789.0 ", value_out, 789);
+    ASSERT_INT_EQUAL_STRING_DECIMAL_TO_LONG("-789.0 ", value_out, (long) -789);
+    ASSERT_INT_EQUAL_STRING_DECIMAL_TO_LONG("9.0           ", value_out, 9);
+    ASSERT_INT_EQUAL_STRING_DECIMAL_TO_LONG("789.0         ", value_out, 789);
+    ASSERT_INT_EQUAL_STRING_DECIMAL_TO_LONG("-789.0        ", value_out, (long) -789);
+
+    // More spaces:
+    ASSERT_INT_EQUAL_STRING_DECIMAL_TO_LONG("   0.0     ", value_out, 0);
+    ASSERT_INT_EQUAL_STRING_DECIMAL_TO_LONG("   -0.0    ", value_out, 0);
+    ASSERT_INT_EQUAL_STRING_DECIMAL_TO_LONG("   123.0  ", value_out, 123);
+
+    // Space separated numbers:
+    ASSERT_INT_EQUAL_STRING_DECIMAL_TO_LONG("   123.456 78   9  ", value_out, 123);
+    ASSERT_INT_EQUAL_STRING_DECIMAL_TO_LONG("123.0 45", value_out, 123);
+
+    // Edge case: it doesnt matter what is after the decimal point
+    ASSERT_INT_EQUAL_STRING_DECIMAL_TO_LONG("789.0b", value_out, 789);
+    ASSERT_INT_EQUAL_STRING_DECIMAL_TO_LONG("123.blah", value_out, 123);
+    ASSERT_INT_EQUAL_STRING_DECIMAL_TO_LONG("456.", value_out, 456);
+}
+
+static void test_string_decimal_to_long_errors(void)
+{
+    // Successful call returns 0
+    long value_out;
+
+    ASSERT_INT_EQUAL_STRING_DECIMAL_TO_LONG("1234.5678", value_out, 1234);
+
+    assert_int_not_equal(StringDecimalToLong("", &value_out), 0);
+    assert_int_not_equal(StringDecimalToLong(" ", &value_out), 0);
+    assert_int_not_equal(StringDecimalToLong("error", &value_out), 0);
+    assert_int_not_equal(StringDecimalToLong("-error", &value_out), 0);
+    assert_int_not_equal(StringDecimalToLong("ffff", &value_out), 0);
+    assert_int_equal(StringDecimalToLong("999999999999999999999999999999.9", &value_out),
+                     -84);
+    assert_int_equal(StringDecimalToLong(" 999999999999999999999999999999.9", &value_out),
+                     -84);
+    assert_int_equal(StringDecimalToLong("-999999999999999999999999999999.9", &value_out),
+                     -84);
+
+    // value_out should be unmodified after errors
+    assert_int_equal(value_out, 1234);
+}
+
 static void test_string_to_ulong(void)
 {
     // Basic usage:
@@ -1424,6 +1494,8 @@ int main()
         unit_test(test_string_from_long),
         unit_test(test_string_to_double),
         unit_test(test_string_from_double),
+        unit_test(test_string_decimal_to_long),
+        unit_test(test_string_decimal_to_long_errors),
 
         unit_test(test_safe_compare),
         unit_test(test_safe_equal),
--- a/libutils/string_lib.h
+++ b/libutils/string_lib.h
@@ -67,6 +67,8 @@ char ToUpper(char ch);
 void ToUpperStrInplace(char *str);
 void ToLowerStrInplace(char *str);
 
+int StringDecimalToLong(const char *str, long *value_out) FUNC_WARN_UNUSED_RESULT;
+
 int StringToLong(const char *str, long *value_out) FUNC_WARN_UNUSED_RESULT;
 void LogStringToLongError(const char *str_attempted, const char *id, int error_code);
 long StringToLongDefaultOnError(const char *str, long default_return);


libutils/string_lib.c
```diff
@@ -528,6 +528,27 @@ int StringToLong(const char *str, long *value_out)
     return 0;
 }
 
+/**
+ * @brief Convert a string representing a real number to int, discarding the decimal part.
+ *
+ * @param[in] str String matching the regex \d+\.\d+, cannot be NULL
```
vpodzime, line 534: If the caller of this function has to assert this first, they don't need this function. It needs to be able to work with at least just `\d+` strings.

vpodzime (commented):
Looks good to me otherwise. Regarding the `int()` policy function, make sure it either supports fancy numbers like `"2k"` that other functions support (and we already have a function to convert strings like that into numbers) or it clearly states that those are not supported.

vpodzime:
> we already have a function to convert strings like that into numbers

`IntFromString()`

libutils/string_lib.c
KotzaBoss: Actually the regex is off. It should be `\d+(\.\d+)?`. Because if only an integer is passed then it still works. The test i have on the other pr confirms it. In essence it scans the string to find the integer part but if there _only_ an integer then it still works since the entire string _is_ the integer part.

KotzaBoss:
ah i see the IntFromString. It has the same fail state when given `"123.456"` the dot messes it up. But now theres a bit of a dilema mostly on semantics. What is the StringDecimalToLong expected to convert? Because the StringToLong doesnt do any fancy "1k" conversion. So the `int` policy function should parse the string like i do in the `StringDecimalToLong` but actually use the `IntFromString` function instead of the one i pred. Am i missing something?

KotzaBoss:
Adding to the semantic confusion... i can just use DoubleFromString and just cast to long to truncate the decimals in the policy function

KotzaBoss:
For the int policy function i updated it to actually use `DoubleFromString` and just discard the decimals. If that is acceptable then the `StringDecimalToLong` is not necessary, unless it could be used somewhere else.

While i proposed this function i am abit confused on the semantics between these two fucntions. the DoubleFromString can be considered a CFEngine logic functions whereas the StringDecimalToLong could be a "C" utility function?

vpodzime:
> the DoubleFromString can be considered a CFEngine logic functions whereas the StringDecimalToLong could be a "C" utility function?

Exactly. Keep in mind that *libntech/libutils* is a general-purpose C library. While it still has some CFEngine-specific aspects, unfortunately, we are trying to eliminate them and make this library as re-usable as possible.

libutils/string_lib.c
```diff
@@ -528,6 +528,27 @@ int StringToLong(const char *str, long *value_out)
...
+int StringDecimalToLong(const char *str, long *value_out)
+{
+    assert(str != NULL);
+    assert(value_out != NULL);
+
+    int len_of_int_part = strcspn(str, ".");
+    char int_part[len_of_int_part];
+
+    strncpy(int_part, str, len_of_int_part);
+    int_part[len_of_int_part] = '\0';
```
vpodzime, line 547: Please make sure this doesn't do extra work if the string is just a simple integer.

libutils/string_lib.c
```diff
@@ -528,6 +528,35 @@ int StringToLong(const char *str, long *value_out)
...
+ * @param[in] str String matching the regex \d+(\.\d+)?, cannot be NULL
+ * @param[out] value_out Where to store result on success, cannot be NULL
+ * @return 0 on success, error code otherwise (see StringToLong source code)
+ */
+int StringDecimalToLong(const char *str, long *value_out)
+{
+    assert(str != NULL);
+    assert(value_out != NULL);
+
+    int len_of_int_part = strcspn(str, ".");
```
vpodzime, line 547: Should be `size_t len_of_int_part`.

libutils/string_lib.c
```diff
@@ -528,6 +528,35 @@ int StringToLong(const char *str, long *value_out)
...
+    if (len_of_int_part < strlen(str))
+    {
+        char int_part[len_of_int_part];
```
vpodzime, line 550: Please add a check that `len_of_int_part` is not beyond some limit, say 128. Otherwise this could easily cause a stack overflow. Feel free to add a new error for this to the list potentially returned by `StringToLong()`.

vpodzime (commented):
Looks good to me otherwise.

libutils/string_lib.c
KotzaBoss: Is 128 a specific value or is there a macro somewhere for a more universal string size?

vpodzime: It's supposed to be a string form of a number, right? So I believe we could even go to 32. There is a macro to determine the maximum length of a number somewhere, so you could use that in combination with `INT_MAX` and double it (for the decimal part) to get the longest size that would work.

KotzaBoss: yes int_part is a string that at most is as long as the input string. So `strlen(int_part) <= strlen(str)` by the end.
But it will potentially hold the integer part of the input string.
```
Input: "123.456"
int_part: "123"

Input: "123456"
int_part: wont be assigned the input will be passed to StringToLong directly
```
Im missing the context behind this number. INT_MAX is a gigantic number how could it help us avoid stack overflow? Arent we looking for a "safe" max size for the array?

vpodzime: Then length of the string representation of `INT_MAX` is the biggest length of the integer part that is supported, right? Actually, `+1` because there might be `-` in front of it. 

So that's the upper bound and it's quite small and safe for stack allocation. Unlike blindly using `strcspn()` on a string where the first `.` can be after arbitrary many bytes and then trying to allocate space for those arbitrary many bytes on stack. The function should be defensive, not rely on code using it with nice and short strings.

libutils/string_lib.c
```diff
@@ -528,6 +528,41 @@ int StringToLong(const char *str, long *value_out)
...
+    assert(value_out != NULL);
+
+    size_t len_of_int_part = strcspn(str, ".");
+
+    if (len_of_int_part > PRINTSIZE(INT_MAX))
+    {
+        return -84;  // Integer part too large to allocate buffer
+    }
+
+    if (len_of_int_part < strlen(str))
```
vpodzime, line 554: `if (str[len_of_int_part] == '\0')` is a cheaper check than using `strlen()` on a potentially long string.

libutils/string_lib.c
```diff
@@ -528,6 +528,41 @@ int StringToLong(const char *str, long *value_out)
...
+int StringDecimalToLong(const char *str, long *value_out)
+{
+    assert(str != NULL);
+    assert(value_out != NULL);
+
+    size_t len_of_int_part = strcspn(str, ".");
+
+    if (len_of_int_part > PRINTSIZE(INT_MAX))
+    {
+        return -84;  // Integer part too large to allocate buffer
```
vpodzime, line 551: Please add handling for this value to `LogStringToLongError()`.

vpodzime (commented):
Looks good to me otherwise.

libutils/string_lib.c
```diff
@@ -529,8 +529,43 @@ int StringToLong(const char *str, long *value_out)
...
+    }
+
+    if (str[len_of_int_part] == '\0')
```
olehermanse, line 554: This condition is opposite? If there is a `\0` there, you don't have to do anything more, just call `StringToLong().`
 
Alternatively, I think it would be even better to flip the ordering here and use an early-return:

```C
if (str[len_of_int_part] == '\0')
{
    return StringToLong(str, value_out);
}
// [code for inserting string terminator and copying int part ...]
```

libutils/string_lib.c
```diff
@@ -529,8 +529,43 @@ int StringToLong(const char *str, long *value_out)
...
+
+        return StringToLong(int_part, value_out);
+    }
+
+    return StringToLong(str, value_out);
+}
+
+/**
+ * @brief Log StringToLong & StringToUlong & StringDecimalToLong conversion error
+ * with an identifier for debugging.
```
olehermanse, line 568: Should still be indented here like before:

```suggestion
 * @brief Log StringToLong & StringToUlong & StringDecimalToLong conversion error
 *        with an identifier for debugging.
```

olehermanse (changes requested):
You should add some unit tests for this to [`string_lib_test.c`](https://github.com/cfengine/libntech/blob/master/tests/unit/string_lib_test.c)

tests/unit/string_lib_test.c
```diff
@@ -496,6 +496,42 @@ static void test_string_to_long_compatibility(void)
     // Old function (StringToLongUnsafe) does not allow trailing whitespace
 }
 
+static void test_string_decimal_to_long(void)
+{
+   // Basic usage:
+   assert_int_equal(StringDecimalToLong("0.0"), 0);
```
olehermanse, line 502: You have to use the output parameter.

tests/unit/string_lib_test.c
```diff
@@ -496,6 +496,42 @@ static void test_string_to_long_compatibility(void)
...
+   assert_int_equal(StringDecimalToLong("-0.0"), 0);
+   assert_int_equal(StringDecimalToLong("+0.0"), 0);
+   assert_int_equal(StringDecimalToLong("123.0"), 123);
```
olehermanse, line 505: `StringToLong()` and `StringDecimalToLong()` return the error code (0), not the result. All the tests are failing because of this. You need to use the output parameter, and check it's value after each call.

```C
   assert_int_equal(StringDecimalToLong("123.0", &value_out), 0);
   assert_int_equal(value_out, 123);
```

(To avoid a lot of repetition, you could put this in a wrapper function, or macro).

tests/unit/string_lib_test.c
```diff
@@ -496,6 +496,42 @@ static void test_string_to_long_compatibility(void)
...
+   assert_int_equals(StringDecimalToLong("-789.0        "), -789);
+
+   // More spaces:
+   assert_int_equals(StringDecimalToLong("   0.0     "), 0);
+   assert_int_equals(StringDecimalToLong("   -0.0    "), 0);
+   assert_int_equals(StringDecimalToLong("   123.0  "), 123);
+
+   // Space separated numbers:
+   assert_int_equals(StringDecimalToLong("   123.456 78   9  "), 123);
+   assert_int_equals(StringDecimalToLong("123.0 45"), 123);
```
olehermanse, line 532: You should add a test case for testing the error handling as well. Send in a too long number and check that it returns nonzero and also check that the string error message is correct.

olehermanse (changes requested):
Implementation looks good now, but test needs more work as it is failing currently.

tests/unit/string_lib_test.c
KotzaBoss: How can i check that the string logged is correct? it is immidiatelly passed to `Log`

olehermanse: Ah, I didn't realize. Then it's not so easy, so drop it.

tests/unit/string_lib_test.c
```diff
@@ -496,6 +497,78 @@ static void test_string_to_long_compatibility(void)
...
+    assert_int_equal(StringDecimalToLong("9999999999999999999999999999999999", &value_out),
+        ERANGE);
+    assert_int_equal(StringDecimalToLong(" 9999999999999999999999999999999999", &value_out),
+        ERANGE);
+    assert_int_equal(StringDecimalToLong("-9999999999999999999999999999999999", &value_out),
+        ERANGE);
+
+    // Can call error logging function
+    LogStringToLongError("-9999999999999999999999999999999999", "string_lib_test",
+                         ERANGE);
```
olehermanse, line 566: Probably drop this.

tests/unit/string_lib_test.c
```diff
@@ -496,6 +497,78 @@ static void test_string_to_long_compatibility(void)
...
+    assert_int_not_equal(StringDecimalToLong("", &value_out), 0);
+    assert_int_not_equal(StringDecimalToLong(" ", &value_out), 0);
+    assert_int_not_equal(StringDecimalToLong("error", &value_out), 0);
+    assert_int_not_equal(StringDecimalToLong("-error", &value_out), 0);
+    assert_int_not_equal(StringDecimalToLong("ffff", &value_out), 0);
+    assert_int_not_equal(StringDecimalToLong("123.456d", &value_out), 0);
+    assert_int_equal(StringDecimalToLong("9999999999999999999999999999999999", &value_out),
+        ERANGE);
+    assert_int_equal(StringDecimalToLong(" 9999999999999999999999999999999999", &value_out),
+        ERANGE);
```
olehermanse, line 560: Test the -84 that you added:

```suggestion
    assert_int_equal(StringDecimalToLong(" 9999999999999999999999999999999999.9", &value_out),
        -84);
```

tests/unit/string_lib_test.c
olehermanse, line 566: Probably drop this.

```suggestion
```

tests/unit/string_lib_test.c
```diff
@@ -496,6 +497,78 @@ static void test_string_to_long_compatibility(void)
...
+    ASSERT_INT_EQUAL_STRING_DECIMAL_TO_LONG("   -0.0    ", 0);
+    ASSERT_INT_EQUAL_STRING_DECIMAL_TO_LONG("   123.0  ", 123);
+
+    // Space separated numbers:
+    ASSERT_INT_EQUAL_STRING_DECIMAL_TO_LONG("   123.456 78   9  ", 123);
+    ASSERT_INT_EQUAL_STRING_DECIMAL_TO_LONG("123.0 45", 123);
+}
+
+#define ASSERT_INT_NOT_EQUAL_STRING_DECIMAL_TO_LONG(str, expected) \
+    assert_int_not_equal(StringDecimalToLong(str, &value_out), 0);
```
olehermanse, line 543: You never use this macro. Also, the name is not good and the `expected` parameter is not used.

```suggestion
#define ASSERT_STRING_DECIMAL_TO_LONG_ERRORS(str)                \
{                                                                \
    long *bad_addr = (long *) 1; /* Should not be written to */  \
    assert_int_not_equal(StringDecimalToLong(str, bad_addr), 0); \
}
```

olehermanse (changes requested):
Some comments on the test only.

libutils/string_lib.c
```diff
@@ -529,8 +529,43 @@ int StringToLong(const char *str, long *value_out)
...
+    }
+
+    if (len_of_int_part == 0 || str[len_of_int_part] == '\0')
+    {
+        return StringToLong(str, value_out);
+    }
+
+    char int_part[len_of_int_part];
+    strncpy(int_part, str, len_of_int_part);
+    int_part[len_of_int_part - 1] = '\0';
```
olehermanse, line 561: So, what is at `str[len_of_int_part]` and `str[len_of_int_part - 1]`?

You should add some asserts and comments about this, because it seems confusing as it is now.

olehermanse (changes requested):

KotzaBoss:
for `StringToLong` the behaviour is to error if the input is something like `1234d`. In the current implentation of StringDecimal it doent really matter what after the `'.'`. Should it scan the input first for validation?
for example:
```
"123.3" ok
"123.4a" not ok
"123.4 a" ok
```

tests/unit/string_lib_test.c
```diff
@@ -496,6 +496,72 @@ static void test_string_to_long_compatibility(void)
     // Old function (StringToLongUnsafe) does not allow trailing whitespace
 }
 
+#define ASSERT_INT_EQUAL_STRING_DECIMAL_TO_LONG(str, expected) \
+    assert_int_equal(StringDecimalToLong(str, &value_out), 0); \
+    assert_int_equal(value_out, expected)
```
olehermanse, line 501: This macro should take value_out as a parameter. It is confusing that you have to declare a value_out in the same scope to use this macro.

tests/unit/string_lib_test.c
```diff
@@ -496,6 +496,72 @@ static void test_string_to_long_compatibility(void)
...
+    long value_out;
+
+    ASSERT_INT_EQUAL_STRING_DECIMAL_TO_LONG("1234.5678", 1234);
+
+    assert_int_not_equal(StringDecimalToLong("", &value_out), 0);
+    assert_int_not_equal(StringDecimalToLong(" ", &value_out), 0);
+    assert_int_not_equal(StringDecimalToLong("error", &value_out), 0);
+    assert_int_not_equal(StringDecimalToLong("-error", &value_out), 0);
+    assert_int_not_equal(StringDecimalToLong("ffff", &value_out), 0);
+    assert_int_not_equal(StringDecimalToLong("123.456d", &value_out), 0);
```
olehermanse, line 553: @KotzaBoss I think it is fine to not do any validation past the `.`. If needed we can have a separate function to validate if something is a floating point number.
```suggestion
```

tests/unit/string_lib_test.c
```diff
@@ -496,6 +496,72 @@ static void test_string_to_long_compatibility(void)
...
+
+    // More spaces:
+    ASSERT_INT_EQUAL_STRING_DECIMAL_TO_LONG("   0.0     ", 0);
+    ASSERT_INT_EQUAL_STRING_DECIMAL_TO_LONG("   -0.0    ", 0);
+    ASSERT_INT_EQUAL_STRING_DECIMAL_TO_LONG("   123.0  ", 123);
+
+    // Space separated numbers:
+    ASSERT_INT_EQUAL_STRING_DECIMAL_TO_LONG("   123.456 78   9  ", 123);
+    ASSERT_INT_EQUAL_STRING_DECIMAL_TO_LONG("123.0 45", 123);
+}
```
olehermanse, line 539:
```suggestion
    // Edge case - it doesn't matter what is after decimal point:
    ASSERT_INT_EQUAL_STRING_DECIMAL_TO_LONG("789.0b", 789);
    ASSERT_INT_EQUAL_STRING_DECIMAL_TO_LONG("123.blah", 123);
    ASSERT_INT_EQUAL_STRING_DECIMAL_TO_LONG("456.", 456);
}
```

tests/unit/string_lib_test.c
olehermanse: Maybe @vpodzime has an opinion on this.

tests/unit/string_lib_test.c
```diff
@@ -496,6 +496,76 @@ static void test_string_to_long_compatibility(void)
     // Old function (StringToLongUnsafe) does not allow trailing whitespace
 }
 
+#define ASSERT_INT_EQUAL_STRING_DECIMAL_TO_LONG(str, value_out, expected) \
+    assert_int_equal(StringDecimalToLong(str, &value_out), 0); \
+    assert_int_equal(value_out, expected)
```
olehermanse, line 501:
```suggestion
    assert_int_equal(value_out, expected);
```

KotzaBoss: the omission of `;` was intentional so that compilation fails.
```c
ASSERT_INT_EQUAL(...)
ASSERT_INT_EQUAL(...)
//becomes
//                           V
assert_int(...); assert_int() assert_int(); assert_int()
``` 
And putting a `;` to mimic a func call now means we have a redundant `;`
```c
ASSERT_INT_EQUAL(...);
// becomes
assert_int(...); assert_int(...);;
```
Isnt the macro without the `;` preferable as it forces a more strict code syntax?

olehermanse: Okay, I see. I think it's more normal to put all the statements inside a scope `{}`. Let's leave it like this, then.

FYI:

Keep in mind that macros are tricky, and `assert_int_equal()` is also a macro.

What is allowed in terms of semicolons and curly brackets depends a lot on the context, for example, consider creating a function-like macro to put here:

```
for (int i=0; i += my_macro(i); ++i)
...
```

olehermanse (approved):

PR #140, opened 2021-05-26
Throw a warning / error when ignoring the return value of StringFormat()

olehermanse:
Ignoring the return value of this function is always a bug:

* You cannot free the string - Memory leak
* It does nothing useful


olehermanse:
@KotzaBoss @larsewi @vpodzime we should use this macro more.

olehermanse:
GitHub Action failure unrelated: https://github.com/cfengine/libntech/pull/141

vpodzime (approved):
ACK

larsewi (commented):
ACK

PR #147, opened 2021-07-22
Add the SeqSoftSet() function

vpodzime:
A function for replacing a value in a sequence without destroying
the replaced value. Useful for moving items away from a sequence
without unnecessary allocations (copying).

Ticket: ENT-5271
Changelog: None

### DIFF
--- a/libutils/sequence.c
+++ b/libutils/sequence.c
@@ -107,6 +107,13 @@ void SeqSet(Seq *seq, size_t index, void *item)
     seq->data[index] = item;
 }
 
+void SeqSoftSet(Seq *seq, size_t index, void *item)
+{
+    assert(seq != NULL);
+    assert(index < SeqLength(seq));
+    seq->data[index] = item;
+}
+
 void SeqAppend(Seq *seq, void *item)
 {
     assert(seq != NULL);
--- a/libutils/sequence.h
+++ b/libutils/sequence.h
@@ -115,7 +115,19 @@ typedef int (*SeqItemComparator) (const void *, const void *, void *user_data);
  */
 int StrCmpWrapper(const void *s1, const void *s2, void *user_data);
 
+/**
+ * Replace value at #index.
+ *
+ * @warning Destroys the original item at #index! See SeqSoftSet().
+ */
 void SeqSet(Seq *set, size_t index, void *item);
+
+/**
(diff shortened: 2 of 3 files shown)


libutils/sequence.c
```diff
@@ -107,6 +107,13 @@ void SeqSet(Seq *seq, size_t index, void *item)
     seq->data[index] = item;
 }
 
+void SeqSoftSet(Seq *seq, size_t index, void *item)
```
KotzaBoss, line 110: Just a wild idea. Would it make sense to actually return the original item? That would allow us to skip using `SeqAt` before setting.
This (like in the [test](https://github.com/cfengine/libntech/pull/147/files#diff-065eb547f1838448271cae05ff88bf1111fde1e9d11708f574a93a9a841b5be1R61-R65))
```c
item = SeqAt(seq, 5);
SeqSoftSet(seq, 5, new_item);
```
could become
```c
item = SeqSoftSet(seq, 5)
```

libutils/sequence.h
```diff
@@ -115,8 +115,20 @@ typedef int (*SeqItemComparator) (const void *, const void *, void *user_data);
...
+ * Replace value at #index.
+ *
+ * @warning Destroys the original item at #index! See SeqSoftSet().
+ */
 void SeqSet(Seq *set, size_t index, void *item);
 
+/**
+ * Set value at #index.
+ *
+ * @note Make sure the original item at #index is destroyed.
```
KotzaBoss, line 128: Adding to my comment above, perhaps we could make sure that we _cant_ discard the return value with a compiler attribute like `FUNC_WARN_UNUSED_RESULT`?
That way we at least know that the compiler will warn us about potential leak.

libutils/sequence.c
vpodzime: > Just a wild idea. Would it make sense to actually return the original item? That would allow us to skip using `SeqAt` before setting.

That could be added as another function like `SeqGetAndSet()`. This is just in the idea of other `Soft` functions that we have.

libutils/sequence.h
vpodzime: Yup, for that new function. :wink:

larsewi (approved):
@KotzaBoss approved this message

PR #149, opened 2021-08-17
Add NULL_JSON() and JSON_NOT_NULL() macros

vpodzime:
To check whether a given JsonElement is/isn't NULL or a JSON
'null' value.

Ticket: ENT-7434
Changelog: None


olehermanse (approved):

larsewi (approved):
ACK

PR #150, opened 2021-08-23
Add an API for logging into a buffer and then committing/discarding it

vpodzime:
For cases where it's not clear whether a bulk of log messages
should be logged or discarded -- e.g. when logging temporary
changes that are either reverted or made persistent.

Ticket: CFE-3708
Changelog: None


vpodzime:
> This pull request **introduces 1 alert** when merging [0083b8d](https://github.com/cfengine/libntech/commit/0083b8ded4b68c7a3ec70f635f58d2a063cf2160) into [f766b1a](https://github.com/cfengine/libntech/commit/f766b1ad68240281ff1e605e1b1f5af67d6ec5f3) - [view on LGTM.com](https://lgtm.com/projects/g/cfengine/libntech/rev/pr-9f798555b37eea7408363a8aa739144f2e48375a)
> 
> **new alerts:**
> 
>     * 1 for Non-constant format string

Don't know how to eliminate this.

libutils/logging.c
```diff
@@ -339,6 +351,26 @@ void VLog(LogLevel level, const char *fmt, va_list ap)
     char *msg = StringVFormat(fmt, ap);
     char *hooked_msg = NULL;
 
+    if (logging_into_buffer &&
+        (level >= min_log_into_buffer_level) && (level <= min_log_into_buffer_level))
```
larsewi, line 355:
```suggestion
        (level >= min_log_into_buffer_level) && (level <= max_log_into_buffer_level))
```
Are you supposed to be comparing to min and max?

libutils/logging.c
```diff
@@ -339,6 +351,26 @@ void VLog(LogLevel level, const char *fmt, va_list ap)
...
+    {
+        assert(log_buffer != NULL);
+
+        if (log_buffer == NULL)
+        {
+            /* Should never happen. */
+            Log(LOG_LEVEL_ERR,
+                "Attempt to log a message to an unitialized log buffer, discarding the message");
```
larsewi, line 363: I think we should consider adding a `LOG_LEVEL_PANIC` or something similar in the logging API, for messages like this. This is so that the user would understand that a bug caused this message, and not the user him/her self. But that is out of scope for this PR.

larsewi (commented):
What happens if multiple threads use this buffer logging feature simultaneously?

libutils/logging.c
vpodzime: Good catch! I would swear I changed the second one to `max`.

libutils/logging.c
vpodzime: Yeah, we have `ProgrammingError()` for those cases, but it aborts the program IIRC. Anyway, that's a bit out of scope of this ticket/PR.

vpodzime:
> What happens if multiple threads use this buffer logging feature simultaneously?

Issues. :smile: But that's why it explicitly says this is not thread-safe. So it's OK for `cf-agent` which is single-threaded, but not good for our daemons, for example.

PR #157, opened 2022-01-26
Add free_array_items() inline function

vpodzime:
No need to write the for loop over and over in all places where
it's needed.

Ticket: ENT-8307
Changelog: None

### DIFF
--- a/libutils/alloc.h
+++ b/libutils/alloc.h
@@ -27,6 +27,7 @@
 
 #include <stdlib.h> // size_t
 #include <stdarg.h> // va_list
+#include <assert.h> // assert()
 
 #include <compiler.h>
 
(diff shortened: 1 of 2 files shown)


libutils/alloc.h
```diff
@@ -39,6 +39,14 @@ void *xmemdup(const void *mem, size_t size);
 int xasprintf(char **strp, const char *fmt, ...) FUNC_ATTR_PRINTF(2, 3);
 int xvasprintf(char **strp, const char *fmt, va_list ap) FUNC_ATTR_PRINTF(2, 0);
 
+static inline void free_array_items(void **array, size_t n_items)
+{
+    for (size_t i = 0; i < n_items; i++)
```
olehermanse, line 44: should add an if or assert for `array == NULL`. The norm is that this should be a no-op, but you might have a good reason for why it should not be allowed.

olehermanse (dismissed):

libutils/alloc.h
vpodzime: It's not free'ing the array itself. So an `assert()` it is. :+1:

larsewi (dismissed):
ACK

larsewi (approved):

PR #164, opened 2022-09-09
Avoid dereferencing a type-punned pointer in passopenfile.cf

vpodzime:
GCC 4.8.5-44 on CentOS 7 complains about this:
```
  passopenfile.c: In function 'PassOpenFile_Put':
  passopenfile.c:404:5: error: dereferencing type-punned pointer will break strict-aliasing rules [-Werror=strict-aliasing]
       *(int*)CMSG_DATA(control_message) = descriptor;
       ^
```

because they have -fstrict-aliasing enabled by default.


olehermanse (approved):

larsewi (approved):
ACK

PR #169, opened 2023-01-02
Fixed MacOS unit tests in GH Actions

olehermanse:
Copied from core: https://github.com/cfengine/core/blob/master/.github/workflows/macos_unit_tests.yml


larsewi (approved):
ACK

larsewi (approved):
ACK

PR #173, opened 2023-01-19
Add unit tests to Github Actions

Lex-2008:
Ticket: ENT-9156


craigcomstock (approved):
cool. I thought I would ask for a ci.yml as in core and such, so we can more easily have dependencies but this looks good and I don't see us needing that anytime soon so no need to do it. :+1:

PR #183, opened 2023-06-02
CFE-4185: Support structured logging in libntech

larsewi:
- Added function to log structured messages
- CFEngine now uses structured logging

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -440,6 +440,18 @@ else
   AM_CONDITIONAL(WITH_PCRE, false)
 fi
 
+dnl systemd structured logging
+
+AC_ARG_WITH([systemd-logging], [AS_HELP_STRING([--with-systemd-logging[[=PATH]]], [support systemd structured logging])], [], [with_systemd_logging=check])
+
+if test "x$with_systemd_logging" != xno
+then
+   CF3_WITH_LIBRARY(systemd_logging, [
+      AC_CHECK_LIB(systemd, sd_journal_sendv, [], [if test "x$with_systemd_logging" != xcheck; then AC_MSG_ERROR(Cannot find systemd library); fi])
+      AC_CHECK_HEADERS(systemd/sd-journal.h, [], [if test "x$with_systemd_logging" != xcheck; then AC_MSG_ERROR(Cannot find systemd headers); fi])
+   ])
+fi
+
 dnl libvirt
 
 AC_ARG_WITH([libvirt],
--- a/libutils/logging.c
+++ b/libutils/logging.c
@@ -346,6 +346,32 @@ void LogToSystemLogStructured(const int level, ...)
 }
 #endif // defined(HAVE_SYSTEMD_SD_JOURNAL_H) && defined(HAVE_LIBSYSTEMD)
 
+void __ImproperUseOfLogToSystemLogStructured(void)
+{
+    // TODO: CFE-4190
+    //  - Make sure CodeQL finds at least the reqired function calls below,
+    //    then remove this code
+    assert(false); // Do not use this function!
+
+    // Required: Missing required "MESSAGE" key.
+    LogToSystemLogStructured(LOG_LEVEL_DEBUG, "FOO", "bogus", "BAR", "doofus");
+
+    // Required: String-literal "MESSAGE" is not the same as "message".
+    LogToSystemLogStructured(LOG_LEVEL_INFO, "FOO", "bogus", "BAR", "doofus", "message", "Hello CFEngine!");
+
+    // Required: Number of format-string arguments is less than number of format specifiers.
+    LogToSystemLogStructured(LOG_LEVEL_ERR, "MESSAGE", "%s %d", "CFEngine");
+
+    // Optional: Number of format-string arguments is greated than number of format specifiers.
+    LogToSystemLogStructured(LOG_LEVEL_ERR, "MESSAGE", "%s %d", "CFEngine", 123, "ROCKS!");
+
+    // Optional: All keys must be uppercase or else they are ignored.
+    LogToSystemLogStructured(LOG_LEVEL_CRIT, "FOO", "bogus", "bar", "DOOFUS", "MESSAGE", "Hello CFEngine!");
+
+    // Optional: Pairs trailing the "MESSAGE" key are ignored.
+    LogToSystemLogStructured(LOG_LEVEL_NOTICE, "FOO", "bogus", "MESSAGE", "Hello CFEngine!", "BAR", "doofus");
+}
+
 #ifndef __MINGW32__
 const char *GetErrorStrFromCode(int error_code)
 {
--- a/libutils/logging.c
+++ b/libutils/logging.c
@@ -29,6 +29,10 @@
 #include <cleanup.h>
 #include <sequence.h>
 
+#if defined(HAVE_SYSTEMD_SD_JOURNAL_H) && defined(HAVE_LIBSYSTEMD)
+#include <systemd/sd-journal.h> /* sd_journal_sendv() */
+#endif // defined(HAVE_SYSTEMD_SD_JOURNAL_H) && defined(HAVE_LIBSYSTEMD)
+
 #include <definitions.h>        /* CF_BUFSIZE */
 
 char VPREFIX[1024] = ""; /* GLOBAL_C */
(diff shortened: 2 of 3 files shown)


larsewi:
@cf-bottom Jenkins with exotics please :)

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/9509/)

(with exotics)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/9509/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-9509/

libutils/logging.c
```diff
@@ -281,6 +285,123 @@ void LogToSystemLog(const char *msg, LogLevel level)
...
+                                      struct iovec *const iov,
+                                      const size_t num_pairs)
+{
+    for (size_t i = 0; i < num_pairs; i++)
+    {
+        const char *const key = va_arg(args, char *);
+        const char *const value = va_arg(args, char *);
+
+        char *const pair = StringFormat("%s=%s", key, value);
+        assert(pair != NULL);
```
vpodzime, line 322: I believe `StringFormat()` never returns `NULL` because it uses `xvasprintf()` that kills the process in such a case, but more `asserts()` never hurt. :)

larsewi:
Rebuild [Build Status](https://ci.cfengine.com/job/pr-pipeline/)

libutils/logging.c
```diff
@@ -281,6 +285,123 @@ void LogToSystemLog(const char *msg, LogLevel level)
...
+void LogToSystemLogStructured(const LogLevel level, ...)
+{
+    va_list args;
+    va_start(args, level);
+
+#ifdef HAVE_SYSTEMD_SD_JOURNAL_H
+    const size_t num_pairs = CountExtraPairs(args);
+    struct iovec iov[num_pairs + 2]; // +2 for "MESSAGE" and "PRIORITY".
+
+    iovecAddExtraPairs(args, &iov[2], num_pairs);
```
vpodzime, line 374: > If ap is passed to a function that uses va_arg(ap,type), then the value of ap is undefined after the return of that function.

Generally I think the safest approach here might be to convert the `args` list into a `Seq` and then work with the sequence.

libutils/logging.h
```diff
@@ -128,6 +128,23 @@ void LoggingSetColor(bool enabled);
  */
 void LogToSystemLog(const char *msg, LogLevel level);
 
+/**
+ * @brief Log a message with structured data to system log.
+ * @details This function takes a variable-length argument list and expects
+ *          structured data passed as key-value pairs, where both keys and
+ *          values are C-strings. Furthermore, the last pair must contain the
+ *          "MESSAGE" key, and the corresponding value accepts a format-string
```
vpodzime, line 136: I wish there was some static assert to ensure that there is actually a `MESSAGE` key as the last pair. Because if it's missing, horrible things can easily happen.

Also, if the value has a format sequence and there are no other arguments, it can easily print whatever piece of memory it finds after the list, right?

Please check what the compiler(s) have to offer and there's nothing useful, we need to write a LGTM check.

vpodzime (commented):
Looks good to me otherwise. Thanks for working on this!

libutils/logging.c
larsewi: Yes, i totally agree, thanks 🚀

libutils/logging.h
larsewi: See ticket [CFE-4190](https://northerntech.atlassian.net/browse/CFE-4190)

olehermanse:
@cf-bottom jenkins, please :)

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/9546/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/9546/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-9546/

olehermanse (approved):

configure.ac
```diff
@@ -440,6 +440,18 @@ else
...
 
+dnl systemd structured logging
+
+AC_ARG_WITH([systemd-logging], [AS_HELP_STRING([--with-systemd-logging[[=PATH]]], [support systemd structured logging])], [], [with_systemd_logging=check])
+
+if test "x$with_systemd_logging" != xno
+then
+   CF3_WITH_LIBRARY(systemd_logging, [
+      AC_CHECK_LIB(systemd, sd_journal_sendv, [SYSTEMD_LDFLAGS="-lsystemd"], [if test "x$with_systemd_logging" != xcheck; then AC_MSG_ERROR(Cannot find systemd library); fi])
+      AC_CHECK_HEADERS(systemd/sd-journal.h, [], [if test "x$with_systemd_logging" != xcheck; then AC_MSG_ERROR(Cannot find systemd headers); fi])
```
vpodzime, line 451: This cannot cause `./configure` to fail, the library is not available everywhere where we build things.

vpodzime (commented):
Looks good to me otherwise.

configure.ac
larsewi: I'm not entirely sure how the `CF3_WITH_LIBRARY` macro works. But for me–when configuring my macOS, which does not have the library–the error is only printed if I specify the `--with-systemd-logging` option to the configure script. I believe how it works when you use `xcheck`, is that it adds the library if it's there by default, but errors if you implicitly specify the `--with-systemd-logging` option. I might be entirely wrong though.

vpodzime: Oh, I missed the built-in `if test "x$with_systemd_logging"` check! I'm used to doing the check after `AC_CHECK_HEADERS()`. Sorry for the noise.

vpodzime (approved):
ACK as long as it passes tests.

larsewi:
@cf-bottom Jenkins please :)

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/9558/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/9558/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-9558/

vpodzime (commented):
Why are we back to not using a sequence?

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/9561/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/9561/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-9561/

vpodzime (dismissed):
ACK

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/9569/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/9569/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-9569/

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/9570/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/9570/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-9570/

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/9578/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/9578/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-9578/

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/9623/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/9623/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-9623/

larsewi:
Seems to be working now. Only selenium failing, but I think this is a known issue. I think what was wrong before was that I was using a variable in the format string. Which potentially could hold `%` characters, causing `StringVFormat` to access memory outside of it's stack frame.

vpodzime (approved):
Looks good to me.

PR #188, opened 2023-08-11
Change license of libntech from GPLv3 to Apache License

vpodzime:
To make libntech usable in non-GPL projects and better suited for wider usage.

Committed without authorship to avoid affecting statistics, these are not line of code changes.

### DIFF
--- a/BSDmakefile
+++ b/BSDmakefile
@@ -3,18 +3,17 @@
 #
 #  This file is part of CFEngine 3 - written and maintained by Northern.tech AS.
 #
-#  This program is free software; you can redistribute it and/or modify it
-#  under the terms of the GNU General Public License as published by the
-#  Free Software Foundation; version 3.
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
 #
-#  This program is distributed in the hope that it will be useful,
-#  but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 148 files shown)


cf-bottom (changes requested):
Please use a company e-mail instead of v***@mykolab.com

vpodzime:
This leaves the following occurrences of GPL in the repo:
```
$ git grep '\(General Public License\|GPL\)'
LICENSE:the GNU General Public License (GPL) v.3.0 or other applicable
libutils/man.c:        ".\\\"under the terms of the GNU General Public License as published by the\n"
libutils/man.c:        ".\\\"GNU General Public License for more details.\n"
libutils/man.c:        ".\\\"You should have received a copy of the GNU General Public License\n"
```
where the former one is the mention of GPL in the COSL license (which we are not changing so the mention has to stay there) and the latter is the code that prints a copyright notice for man pages which is currently only used in CFEngine which is GPL licensed and so the license there is correct. Ideally, the code should be able to use a different license, but that's out of scope here.

kacf (approved):
A couple of things I spotted:

* The commercial license at the bottom of `LICENSE` still mentions GPL as the alternative.

* `libutils/man.c` still has GPL in its copyright C string.

Everything else looks good to me!

vpodzime:
> A couple of things I spotted:

Yes, I mentioned both [above](https://github.com/cfengine/libntech/pull/188#issuecomment-1674598691)

kacf:
Ah right, right, I scrolled past that too quickly. All good then.

olehermanse (approved):

PR #196, opened 2023-10-10
CFE-3991: Added CompareVersionExpression() for comparing 2 version numbers

olehermanse:
Will be used in the upcoming CFEngine policy function; `version_compare()`.

This function takes 3 arguments, 2 version numbers and an
operator between them, thus resembling an expression to evaluate.
(Is this version bigger than this version, etc.)

Internally it uses the exact same logic as we use for comparing the
currently installed CFEngine version, for the cf_version_minimum
and similar functions and macros.

### DIFF
--- a/tests/unit/version_comparison_test.c
+++ b/tests/unit/version_comparison_test.c
@@ -50,12 +50,33 @@ static void test_CompareVersion(void)
     assert_true(VERSION_ERROR == CompareVersion("", "3.16.0"));
 }
 
+static void test_CompareVersionExpression(void)
+{
+    assert_true(BOOLEAN_TRUE == CompareVersionExpression("1.2.3", "=", "1.2.3"));
+    assert_true(BOOLEAN_TRUE == CompareVersionExpression("1.2.3", "==", "1.2.3"));
+    assert_true(BOOLEAN_TRUE == CompareVersionExpression("1.2.3", "!=", "1.2.4"));
+    assert_true(BOOLEAN_TRUE == CompareVersionExpression("100.0.0", ">", "99.0.0"));
+    assert_true(BOOLEAN_TRUE == CompareVersionExpression("100.0.0", ">=", "99.0.0"));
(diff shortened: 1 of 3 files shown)


tests/unit/version_comparison_test.c
```diff
@@ -50,12 +50,33 @@ static void test_CompareVersion(void)
     assert_true(VERSION_ERROR == CompareVersion("", "3.16.0"));
 }
 
+static void test_CompareVersion(void)
```
nickanderson, line 53: I don't see any comparisons here that include fun suffixes.

e.g. pulled from random entries from `apt list`

-   `0.13.72+dfsg.1-1.1`
-   `1.0.0+git20190521.78d0385+1.8.2-2build5`
-   `22.04.3-0ubuntu1~ubuntu22.04~ppa1`

What would the expected behavior of using something like that?

nickanderson (commented):
Can't really comment on the code, but I do have a question about more interesting version strings, e.g. from package managers etc ...

tests/unit/version_comparison_test.c
olehermanse: @nickanderson I haven't changed any of the version comparison logic - it's the same used for `cf_version_minimum` and friends. I believe everything after patch is ignored so `1.2.3`, `1.2.3-1`, `1.2.3-2`, `1.2.3+git1234`, `1.2.3a` are all considered the same version.

This is intentional and often desirable. It's very messy to find a generic solution to comparing the rest of a version string, and if we'd like to attempt that, I think it should be in a separate function. (i.e. something more specialized `compare_version_with_release`, `compare_version_git`).

In policy you can combine version_compare and strcmp to get more info if desirable:

```
bundle agent main
{
  reports:
    "Same major.minor.patch version but not exactly the same version string"
      if => and(version_compare("$(a)", "=", "$(b)"), not(strcmp("$(a)", "$(b)")
}
```

vpodzime (approved):
ACK

larsewi (approved):

PR #197, opened 2023-11-03
Change MapSize() complexity to O(1) even if it's a hash map

vpodzime:
No need to iterate, a HashMap has the `load` field which contains the number of elements in the map.

Ticket: CFE-3043
Changelog: Complexity of MapSize() is now O(1) in all cases

### DIFF
--- a/libutils/map.c
+++ b/libutils/map.c
@@ -118,15 +118,7 @@ size_t MapSize(const Map *map)
     }
     else
     {
-        MapIterator i = MapIteratorInit((Map*)map);
-        size_t size = 0;
-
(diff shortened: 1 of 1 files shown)


olehermanse (approved):
Next step, make it a `static inline` function?

vpodzime:
> Next step, make it a `static inline` function?

It's definitely an option. Although, I think the stuff in `map.h` (where it would have to be defined) should use as least of the `_priv.h` headers as possible. In fact, we should try to remove their `#include`s from `map.h`.

larsewi (approved):

PR #203, opened 2023-12-08
Added new StringJoin() utility function

larsewi:
The function joins a sequence of strings with an optional separator in between the elements.

### DIFF
--- a/libutils/string_lib.c
+++ b/libutils/string_lib.c
@@ -1781,3 +1781,23 @@ bool StringMatchesOption(
     }
     return StringEqualN_IgnoreCase(supplied, longopt, length);
 }
+
+char *StringJoin(const Seq *const seq, const char *sep)
+{
+    assert(seq != NULL);
+
+    Writer *const writer = StringWriter();
+    const size_t len = SeqLength(seq);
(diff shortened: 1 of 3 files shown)


craigcomstock (approved):
the code looks so familiar to something I saw recently! ;) 🏆

libutils/string_lib.c
```diff
@@ -1781,3 +1781,23 @@ bool StringMatchesOption(
...
     return StringEqualN_IgnoreCase(supplied, longopt, length);
 }
+
+char *StringJoin(const Seq *const seq, const char *sep)
+{
+    assert(seq != NULL);
+
+    Writer *const writer = StringWriter();
+    const size_t len = SeqLength(seq);
+    for (size_t i = 0; i < len; i++)
```
vpodzime, line 1791: I'd do `i < (len -1)` writing the separator *after* every item and then writing the last item after the `for` loop without a separator. But YMMV.

vpodzime (approved):
Looks good to me otherwise.

libutils/string_lib.c
vpodzime: Yup, but that's easy to check upfront.

PR #204, opened 2023-12-11
Define type Regex and implement RegexDestroy()

vpodzime:
When migrating from PCRE to PCRE2 we had to make a change breaking API because all the functions working with REs were using the PCRE's type `pcre`. To avoid this change in the future, we define our own type `Regex` which we can later redefine without causing any API breakage.

This also requires a definition of `RegexDestroy()` for use instead of `pcre2_code_free()`.

Ticket: ENT-10629
Changelog: API working with regular expressions uses a custom
           type Regex instead of pcre2_code directly

### DIFF
--- a/libutils/json-pcre.h
+++ b/libutils/json-pcre.h
@@ -8,6 +8,6 @@
 #include <regex.h>
 
 JsonElement *StringCaptureData(
-    const pcre2_code *regex, const char *pattern, const char *data);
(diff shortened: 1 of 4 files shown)


craigcomstock (approved):
looks great! thanks!

olehermanse (approved):
Cool, thanks!

PR #217, opened 2024-04-17
Fixed JsonCompare() function and switched to using it for json_test

craigcomstock:
Ticket: ENT-11450
Changelog: none

### DIFF
--- a/libutils/json.c
+++ b/libutils/json.c
@@ -76,6 +76,20 @@ struct JsonElement_
 // JsonElement Functions
 // *******************************************************************************************
 
+const char *JsonContainerTypeToString(const JsonContainerType type)
+{
+    switch (type)
+    {
+    case JSON_CONTAINER_TYPE_ARRAY:
+        return "array";
+    case JSON_CONTAINER_TYPE_OBJECT:
+        return "object";
+    default:
+        UnexpectedError("Unknown JSON container type: %d", type);
+        return "(null)";
+    }
+}
+
 const char *JsonPrimitiveTypeToString(const JsonPrimitiveType type)
 {
     switch (type)
@@ -262,11 +276,12 @@ static int JsonArrayCompare(
     int ret = JsonLength(a) - JsonLength(b);
     if (ret != 0)
     {
+        Log(LOG_LEVEL_DEBUG, "JsonArrayCompare() fails, length differs by %d", ret);
         return ret;
     }
 
     JsonIterator iter_a = JsonIteratorInit(a);
-    JsonIterator iter_b = JsonIteratorInit(a);
+    JsonIterator iter_b = JsonIteratorInit(b);
 
     for (size_t i = 0; i < JsonLength(a); i++)
     {
@@ -276,6 +291,7 @@ static int JsonArrayCompare(
         ret = JsonCompare(child_a, child_b);
         if (ret != 0)
         {
+            Log(LOG_LEVEL_DEBUG, "JsonArrayCompare() fails for index %zu, children not equal", i);
             return ret;
         }
     }
@@ -296,34 +312,32 @@ static int JsonObjectCompare(
     int ret = JsonLength(a) - JsonLength(b);
     if (ret != 0)
     {
+        Log(LOG_LEVEL_DEBUG, "JsonObjectCompare() fails, length differs by %d", ret);
         return ret;
     }
 
-    JsonIterator iter_a = JsonIteratorInit(a);
-    JsonIterator iter_b = JsonIteratorInit(a);
+    JsonIterator iter = JsonIteratorInit(a);
 
-    for (size_t i = 0; i < JsonLength(a); i++)
+    while (JsonIteratorHasMore(&iter))
     {
-        const JsonElement *child_a = JsonIteratorNextValue(&iter_a);
-        const JsonElement *child_b = JsonIteratorNextValue(&iter_b);
+        const char *const key = JsonIteratorNextKey(&iter);
+        const JsonElement *const child_a = JsonIteratorCurrentValue(&iter);
+        const JsonElement *const child_b = JsonObjectGet(b, key);
 
-        const char *const key_a = JsonIteratorCurrentKey(&iter_a);
-        const char *const key_b = JsonIteratorCurrentKey(&iter_b);
-
-        ret = strcmp(key_a, key_b);
-        if (ret != 0)
+        if (child_b == NULL)
         {
-            return ret;
+            Log(LOG_LEVEL_DEBUG, "JsonObjectCompare() fails for key '%s', not present in object b", key);
+            return 1;
         }
-
         ret = JsonCompare(child_a, child_b);
         if (ret != 0)
         {
+            Log(LOG_LEVEL_DEBUG, "JsonObjectCompare() fails for key '%s', children are not equal", key);
             return ret;
         }
     }
 
-    return ret;
+    return ret; // here ret will be 0 in the case that both objects are empty or all children are equal
 }
 
 
@@ -340,6 +354,7 @@ static int JsonContainerCompare(
 
     if (type_a != type_b)
     {
+        Log(LOG_LEVEL_DEBUG, "JsonContainerCompare() fails, container type '%s' not equal to container type '%s'", JsonContainerTypeToString(type_a), JsonContainerTypeToString(type_b));
         return type_a - type_b;
     }
 
@@ -367,6 +382,7 @@ int JsonCompare(const JsonElement *const a, const JsonElement *const b)
 
     if (type_a != type_b)
     {
+        Log(LOG_LEVEL_DEBUG, "JsonCompare() fails, type %d not equal to type %d", type_a, type_b);
         return type_a - type_b;
     }
 
@@ -376,7 +392,14 @@ int JsonCompare(const JsonElement *const a, const JsonElement *const b)
         return JsonContainerCompare(a, b);
 
     case JSON_ELEMENT_TYPE_PRIMITIVE:
-        return strcmp(a->primitive.value, b->primitive.value);
+    {
+        if (!StringEqual(a->primitive.value, b->primitive.value))
+        {
+            Log(LOG_LEVEL_DEBUG, "JsonCompare() fails, primitive '%s' not equal to '%s'", a->primitive.value, b->primitive.value);
+            return 1;
+        }
+        return 0;
+    }
 
     default:
         UnexpectedError("Unknown JSON element type: %d", type_a);
@@ -1198,7 +1221,7 @@ static int JsonElementHasProperty(
 
     assert(element->propertyName != NULL);
 
-    if (strcmp(propertyName, element->propertyName) == 0)
+    if (StringEqual(propertyName, element->propertyName))
     {
         return 0;
     }
--- a/tests/unit/json_test.c
+++ b/tests/unit/json_test.c
@@ -1,4 +1,5 @@
 #include <test.h>
+#include <logging.h>
 
 #include <json.h>
 #include <string_lib.h>
@@ -567,6 +568,25 @@ static void test_copy_compare(void)
 
     JsonDestroy(bench);
     JsonDestroy(copy);
+}
+
+static void test_compare_container_type_mismatch(void)
+{
+    JsonElement *object_a = JsonObjectCreate(1);
+    JsonElement *object_b = JsonObjectCreate(1);
+
+    JsonElement *child = JsonObjectCreate(1);
+    JsonObjectAppendObject(object_a, "key", child);
+
+    JsonElement *array = JsonArrayCreate(1);
+    JsonArrayAppendString(array, "first");
+    JsonArrayAppendString(array, "second");
+    JsonObjectAppendArray(object_b, "key", array);
+
+    assert_true(JsonCompare(object_a, object_b) != 0);
+
+    JsonDestroy(object_a);
+    JsonDestroy(object_b);
 }
 
 static void test_select(void)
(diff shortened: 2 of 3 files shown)


tests/unit/test.h
```diff
@@ -55,4 +56,13 @@ void test_progress_end(void);
     }\
 }\
 
+// shared json test helpers, used in nova/unit/tests for cf-reactor
```
olehermanse, line 59: @craigcomstock @larsewi @vpodzime Instead of exposing these in a `test.h`, how about just making them a part of the library? It could be a separate file `libntech/libutils/json_comparison.h`. Comparing JSONs could be usable in other places, as well.

vpodzime: I'd expect us to already have that, actually.

larsewi: We do have
```C
int JsonCompare(const JsonElement *a, const JsonElement *b);
```
in `json.h`

larsewi:
```C
    /* JsonCompare is naive and does not handle containers properly, thus we
     * need a custom comparison function to test for equality. */
```
I guess that is why there is a custom one in `json_test.c`? I have a feeling that I wrote this code once upon a time.

craigcomstock: I compared the JsonCompare in json.c with that in json_test.c and didn't see any functional difference so I guess we just don't need those and can use JsonCompare as-is. Tests pass here for me locally so will see how github likes it.

craigcomstock:
I added a negative test and debug logging to see how that might look in case someone wants to debug WHY JsonCompare() found two json's to not be equal

```
test_json_object_merge_deep: Starting test
   debug: JsonCompare() fails, primitive '/tmp/foo' not equal to '/tmp/foobad'
   debug: JsonArrayCompare() fails, children not equal
   debug: JsonObjectCompare(), for key 'value', child objects not equal
   debug: JsonObjectCompare(), for key 'cfbs:delete_files.filenames_bad', child objects not equal
   debug: JsonObjectCompare(), for key 'variables', child objects not equal
merged json  : {"variables":{"cfbs:delete_files.filenames_bad":{"value":["/tmp/foo","/tmp/bar","/tmp/bar","/tmp/baz"]}}}
expected json: {"variables":{"cfbs:delete_files.filenames_bad":{"value":["/tmp/foobad","/tmp/bar","/tmp/bar","/tmp/baz"]}}}
test_json_object_merge_deep: Test completed successfully.
```

The output is a bit "backwards" by noting the inner-most problem and then the elements outside of that in sequence to furthest out last.

libutils/json.c
```diff
@@ -296,31 +298,24 @@ static int JsonObjectCompare(
...
+        const JsonElement *const child_a = JsonObjectGet(a, key);
+        const JsonElement *const child_b = JsonObjectGet(b, key);
 
-        ret = JsonCompare(child_a, child_b);
-        if (ret != 0)
+        if (child_b == NULL || JsonCompare(child_a, child_b))
         {
-            return ret;
+            Log(LOG_LEVEL_DEBUG, "JsonObjectCompare() fails for key '%s', children not equal", key);
+            return 1;
```
craigcomstock, line 316: I feel like just to be sure saying 
```suggestion
            return 1; // not equal like strcmp()
```

libutils/json.c
```diff
@@ -296,31 +298,24 @@ static int JsonObjectCompare(
...
         }
+        return 0;
```
craigcomstock, line 318:
```suggestion
        return 0; // equal, like strcmp()
```

libutils/json.c
```diff
@@ -340,6 +335,7 @@ static int JsonContainerCompare(
 
     if (type_a != type_b)
     {
+        Log(LOG_LEVEL_DEBUG, "JsonContainerCompare() fails, type %d not equal to type %d", type_a, type_b);
```
larsewi, line 338: You may consider logging the string representation of the type, e.g., `null`, `true`, `false`, `number`, `object`, `array`. But not strictly necessary in a debug log message in my opinion.

libutils/json.c
```diff
@@ -376,7 +374,10 @@ int JsonCompare(const JsonElement *const a, const JsonElement *const b)
         return JsonContainerCompare(a, b);
 
     case JSON_ELEMENT_TYPE_PRIMITIVE:
-        return strcmp(a->primitive.value, b->primitive.value);
+        ret = strcmp(a->primitive.value, b->primitive.value);
+        if (ret)
+            Log(LOG_LEVEL_DEBUG, "JsonCompare() fails, primitive '%s' not equal to '%s'", a->primitive.value, b->primitive.value);
+        return ret;
```
larsewi, line 380: We normally declare the variable where it's used. In the case of switch statements, this normally issues a warning unless the variable is inside a new scope. Because it is potentially dangerous to have a conditionally declared variable in the outer scope.
```suggestion
        {
            int ret = strcmp(a->primitive.value, b->primitive.value);
            if (ret)
                Log(LOG_LEVEL_DEBUG, "JsonCompare() fails, primitive '%s' not equal to '%s'", a->primitive.value, b->primitive.value);
            return ret;
        }
```
If you accept this suggestion, also remove the declaration above :)

tests/unit/json_test.c
```diff
@@ -1908,116 +1909,92 @@ static void test_json_null_not_null(void)
...
     }
 
-    JsonDestroy(base);
-    JsonDestroy(extra);
-    JsonDestroy(expected);
+    finish:
+        if (actual_string)
+        {
+            if (pass == false)
+                printf("merged json  : %s\n", actual_string);
```
larsewi, line 1978: > *Always* use curly braces
   - Ref [CONTRIBUTING.md](https://github.com/cfengine/core/blob/master/CONTRIBUTING.md#readability--maintainability)

tests/unit/json_test.c
```diff
@@ -1908,116 +1909,92 @@ static void test_json_null_not_null(void)
...
+        pass = false;
+        goto finish;
     }
 
-    /* JsonCompare is naive and does not handle containers properly, thus we
-     * need a custom comparison function to test for equality. */
-    if (!compare_json_object_merge_deep(actual, expected))
+    actual_string = JsonToString(actual);
+
+    if (JsonCompare(actual, expected))
```
larsewi, line 1951: Use explicit comparison
```suggestion
    if (JsonCompare(actual, expected) != 0)
```

libutils/json.c
olehermanse: Also, use explicit comparisons for integers (`if (ret != 0)`).

libutils/json.c
```diff
@@ -296,31 +298,24 @@ static int JsonObjectCompare(
...
+        const char *const key = JsonIteratorNextKey(&iter);
+        const JsonElement *const child_a = JsonObjectGet(a, key);
+        const JsonElement *const child_b = JsonObjectGet(b, key);
 
-        ret = JsonCompare(child_a, child_b);
-        if (ret != 0)
+        if (child_b == NULL || JsonCompare(child_a, child_b))
         {
-            return ret;
+            Log(LOG_LEVEL_DEBUG, "JsonObjectCompare() fails for key '%s', children not equal", key);
```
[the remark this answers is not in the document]
craigcomstock: correct, will fix.

tests/unit/json_test.c
craigcomstock: I suspected as much as I wrote it but couldn't help myself ;)

libutils/json.c
craigcomstock: This return 0 inside the loop is actually rather wrong... it would cause us to only check the first child element of each object I think. I will add a test to check the code as-is.

craigcomstock: fixed, sorry about the json formatting at the end (hard to read but there)

```
   debug: JsonCompare() fails, primitive 'due' not equal to 'dva'
   debug: JsonObjectCompare() fails for key 'two', children are not equal
   debug: JsonObjectCompare() fails for key 'badchild', children are not equal
   debug: JsonObjectCompare() fails for key 'cfbs:delete_files.filenames_bad_two', children are not equal
   debug: JsonObjectCompare() fails for key 'variables_bad_two', children are not equal
merged json  : {"variables_bad_two":{"cfbs:delete_files.filenames_bad_two":{"badchild":{"one":"uno","two":"due"},"value":["/tmp/foo","/tmp/bar","/tmp/bar","/tmp/baz"]}}}
expected json: {"variables_bad_two":{"cfbs:delete_files.filenames_bad_two":{"badchild":{"one":"uno","two":"dva"},"value":["/tmp/foo","/tmp/bar","/tmp/bar","/tmp/baz"]}}}
```

libutils/json.c
craigcomstock: This is the container case, so should only be array or object right? I can add JsonContainerTypeToString() and use that...

craigcomstock: got it

```
test_compare_container_type_mismatch: Starting test
   debug: JsonContainerCompare() fails, container type 'object' not equal to container type 'array'
   debug: JsonObjectCompare() fails for key 'key', children are not equal
```

libutils/json.c
```diff
@@ -296,34 +312,32 @@ static int JsonObjectCompare(
...
-    for (size_t i = 0; i < JsonLength(a); i++)
+    while (JsonIteratorHasMore(&iter))
     {
-        const JsonElement *child_a = JsonIteratorNextValue(&iter_a);
-        const JsonElement *child_b = JsonIteratorNextValue(&iter_b);
-
-        const char *const key_a = JsonIteratorCurrentKey(&iter_a);
-        const char *const key_b = JsonIteratorCurrentKey(&iter_b);
+        const char *const key = JsonIteratorNextKey(&iter);
+        const JsonElement *const child_a = JsonObjectGet(a, key);
```
vpodzime, line 324: This could use `JsonIteratorCurrentValue()` should be cheaper. But not a biggie.

libutils/json.c
```diff
@@ -376,7 +392,12 @@ int JsonCompare(const JsonElement *const a, const JsonElement *const b)
         return JsonContainerCompare(a, b);
 
     case JSON_ELEMENT_TYPE_PRIMITIVE:
-        return strcmp(a->primitive.value, b->primitive.value);
+    {
+        int ret = strcmp(a->primitive.value, b->primitive.value);
```
vpodzime, line 396: `StringEqual()`

libutils/json.c
```diff
@@ -76,6 +76,20 @@ struct JsonElement_
 // JsonElement Functions
 // *******************************************************************************************
 
+const char *JsonContainerTypeToString(const JsonContainerType type)
```
vpodzime, line 79: I'd make this `static inline` and moved it to the header, but not a biggie.

tests/unit/json_test.c
```diff
@@ -1908,116 +1928,107 @@ static void test_json_null_not_null(void)
...
+            if (pass == false)
+            {
+                printf("expected json: %s\n", expected_string);
+            }
+            free(expected_string);
+        }
+        // in second step of JsonObjectMergeDeepInplace() actual should point to base
+        if (actual != base && actual != NULL)
+        {
+            JsonDestroy(actual);
```
vpodzime, line 2016: `JsonDestroy()` is safe to call with `NULL` so I believe this call and the ones below can be unconditional. **But it's still important to make sure it's not called twice on the same pointer.** (which `!= NULL` doesn't guarantee, of course)

vpodzime (commented):
Very nice fixes and improvements! Thanks!

libutils/json.c
craigcomstock: cargo cult copy paste, want me to fix the other instances in the file? ;) I think there are some.

libutils/json.c
craigcomstock: again, doing it as-it-was-already with other *TypeToString() functions.

libutils/json.c
craigcomstock: What if the keys were out-of-order between element a and b though? I think this should stay as-is.

libutils/json.c
craigcomstock: Looking it would seem there are MANY functions which could be moved to the header and made `static inline`. Not sure where/if we should start/stop. @vpodzime lemme know what you think... just look through the first part of the file. e.g.

```
127	const char *JsonElementGetPropertyName(const JsonElement *const element)
128	{
129	    assert(element != NULL);
130	
131	    return element->propertyName;
132	}
```

tests/unit/json_test.c
craigcomstock: I was trying to be careful above since we re-use the actual pointer I set it to NULL so that this later code "knows" that it has been destroyed. Let me know what you think of this strategy or suggestions on a different way to go. I think I initially added these checks to avoid a double-free and why I set to NULL earlier.

line 1980: (cool line number ;)
```
    JsonDestroy(actual);
    actual = NULL;
    actual = JsonObjectMergeDeepInplace(base, extra);
```

olehermanse: @craigcomstock we have the macro `DESTROY_AND_NULL` from `alloc.h` for this;

```
DESTROY_AND_NULL(JsonDestroy, actual);
```

libutils/json.c
olehermanse: @craigcomstock You'd need to use `JsonObjectGet()` for `b`, but for `a`, you can use `JsonIteratorCurrentValue()` - it will give you the value corresponding to `key`, which you just got from `JsonIteratorNextKey()`.

olehermanse (approved):

libutils/json.c
craigcomstock: Right. Could just iterate over `a` , use `JsonIteratorCurrentValue()` and then check `JsonObjectGet()` for `b`.

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/10628/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/10628/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-10628/

libutils/json.c
vpodzime: Exactly. It's just one less lookup.

libutils/json.c
```diff
@@ -1198,7 +1221,7 @@ static int JsonElementHasProperty(
 
     assert(element->propertyName != NULL);
 
-    if (strcmp(propertyName, element->propertyName) == 0)
+    if (StringEqual(propertyName, element->propertyName))
```
vpodzime, line 1224: :+1:

tests/unit/json_test.c
```diff
@@ -1908,116 +1928,106 @@ static void test_json_null_not_null(void)
...
+
+    if (JsonCompare(actual, expected) != 0)
     {
-        return false;
+        pass = false;
+        goto finish;
     }
 
-    JsonDestroy(actual);
+    DESTROY_AND_NULL(JsonDestroy,actual);
```
vpodzime, line 1979: :+1:

tests/unit/json_test.c
```diff
@@ -1908,116 +1928,106 @@ static void test_json_null_not_null(void)
...
+            JsonDestroy(base);
+        }
+        if (extra != NULL)
+        {
+            JsonDestroy(extra);
+        }
+        if (expected != NULL)
+        {
+            JsonDestroy(expected);
+        }
```
vpodzime, line 2028: Still think these checks for `!= NULL` should not be necessary, but they don't hurt.

vpodzime (approved):
I'd do two things differently, but it's definitely good to go as is.

tests/unit/json_test.c
olehermanse: I agree, also since as a reader it makes it look like it's necessary to check for `NULL`. To be extra clear, you could do:

```
JsonDestroy(expected); // Can be NULL / no-op
```

olehermanse (approved):
LGTM, but take a look at Jenkins test failures before merging.

libutils/json.c
craigcomstock: changed. thanks! :)

tests/unit/json_test.c
craigcomstock: yep, changed. thanks!

craigcomstock:
@cf-bottom jenkins

craigcomstock:
> LGTM, but take a look at Jenkins test failures before merging.

One was a 52 exit code during cfbs test which is likely "known" and should already be addressed so maybe the fix for https://northerntech.atlassian.net/browse/ENT-11526 (apache graceful restart) didn't quite fix it.

The other is maybe new, I have logged a ticket: https://northerntech.atlassian.net/browse/ENT-11565

craigcomstock:
PR to pull this change (and any others up to this merge) into core https://github.com/cfengine/core/pull/5483

PR #223, opened 2024-07-12
Make `array` & `new_array` const-correct in `test_is_string_in_array`.

sp1ff:
When building on Arch, I saw a few compilation errors of the form:

```
string_lib_test.c:873:40: error: passing argument 2 of ‘IsStringInArray’ from incompatible pointer type [-Wincompatible-pointer-types]
```

This small commit fixes them.

### DIFF
--- a/tests/unit/string_lib_test.c
+++ b/tests/unit/string_lib_test.c
@@ -869,7 +869,7 @@ static void test_safe_equal_n(void)
 
 static void test_is_string_in_array(void)
 {
-    char *array[5] = {"one", "two", "three", "four", "five"};
+    const char *array[5] = {"one", "two", "three", "four", "five"};
     assert_true(IsStringInArray("one", array, 5));
     assert_true(IsStringInArray("two", array, 5));
     assert_true(IsStringInArray("three", array, 5));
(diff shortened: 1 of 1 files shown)


larsewi (dismissed):
Thanks 🚀 

One note:
Since the `new_array` variable is allocated on the heap and later free'd, it may be better to keep it `char **` and instead explicitly cast it to `const char **` when passing it as an argument to `IsStringInArray`. The reason is that `free(3)` takes a `void *` as an argument (not `const void *`).

olehermanse:
> Thanks 🚀
> 
> One note: Since the `new_array` variable is allocated on the heap and later free'd, it may be better to keep it `char **` and instead explicitly cast it to `const char **` when passing it as an argument to `IsStringInArray`. The reason is that `free(3)` takes a `void *` as an argument (not `const void *`).

No, how he did it is correct. This is a double pointer, `strings` is a non-const pointer, to a non-const pointer, to a `const char` in memory.

```
#include <stdio.h> // printf()                                                                                                                               
#include <stdlib.h> // calloc()                                                                                                                              

int main(void)
{
  const char **strings = calloc(3, sizeof(const char *));

  // C compiler is not very strict with string literals                                                                                                      
  strings[0] = "hello";	// This works regardless of whether strings has const or not                                                                         

  // After assigning to a typed variable, th compiler willbe more strict:                                                                                    
  const char *world = "world";
  strings[1] = world; // Const correctness is checked here.                                                                                                  
  printf("%s, %s\n", strings[0], strings[1]);

  void *temp = strings;
  // const void *temp = strings; // This would not work, const mismatch.                                                                                     

  free(temp); // free(strings) would also work                                                                                                               
  return 0;
}
```

```
$ gcc -Wall -Werror -Wpedantic test.c -o test && ./test
hello, world
```

sp1ff:
Thanks-- works much better!

PR #224, opened 2024-07-16
Added PathExists() convenience function

olehermanse:


libutils/file_lib.h
```diff
@@ -67,6 +67,16 @@ Writer *FileReadFromFd(int fd, size_t size_max, bool *truncated);
 
 bool FileCanOpen(const char *path, const char *modes);
 
+/**
+ * Returns true if a path (such as a file or a folder) exists
+ * Don't follow symlinks, don't check permissions, don't check if it's a file.
```
vpodzime, line 72: This is very similar to `access(path, F_OK)`, the only difference is that `access()` follows symlinks. Maybe it could be mentioned here as an alternative for such cases?

vpodzime (approved):
ACK

libutils/file_lib.h
olehermanse: Yes, good idea.

larsewi (approved):

craigcomstock (dismissed):
There are about 40 cases in core which check for an error, rc == -1, with lstat and a few that do this exact check. Seems like a solid addition. :+1:

larsewi (approved):

vpodzime (approved):

craigcomstock (approved):

PR #226, opened 2024-08-05
Fixed compiler warning for undefined reference of lstat

larsewi:
Ticket: None
Changelog: None


nickanderson (approved):
Seems sensible and not dangerous. But I am not well qualified to provide this

PR #229, opened 2024-10-08
Created function to copy file while keeping perms

larsewi:


vpodzime (commented):
We already have [`CopyRegularFileDisk()`](https://github.com/cfengine/core/blob/master/libpromises/files_copy.c#L106) in *libpromises* which does the same and in a better/more advanced way AFAICT.

larsewi:
> We already have [CopyRegularFileDisk()](https://github.com/cfengine/core/blob/master/libpromises/files_copy.c#L106) in libpromises which does the same and in a better/more advanced way AFAICT.

Thanks :beers:

PR #230, opened 2024-11-21
Fixed compiler warning -Werror=enum-int-mismatch

larsewi:
Ticket: None
Changelog: None

### DIFF
--- a/libutils/logging.h
+++ b/libutils/logging.h
@@ -141,10 +141,18 @@ void LogToSystemLog(const char *msg, LogLevel level);
  *          log level argument and added to the structured log message.
  * @note Only the formatted string from the pair with the "MESSAGE" key is
  *       logged on platforms that does not support structured logging.
+ *
+ *       The function takes an int as the log level, although the desired type
+ *       would be the enum "LogLevel". For some C compilers, the last required
+ *       argument must be self-promoting: that is, the default promotions must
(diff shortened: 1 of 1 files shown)


larsewi:
@cf-bottom jenkins please :)

libutils/logging.h
```diff
@@ -141,10 +141,16 @@ void LogToSystemLog(const char *msg, LogLevel level);
  *          log level argument and added to the structured log message.
  * @note Only the formatted string from the pair with the "MESSAGE" key is
  *       logged on platforms that does not support structured logging.
+ *
+ *       The function takes an int as the log level, although the deired type
+ *       is LogLevel. However, it is not safe to use an enum next to a
+ *       variable length argument list. Hence, we must use a type of a fixed
```
craigcomstock, line 147: it is not safe, as I understand from a bit of reading, because the underlying type of enums is undefined/implementation specific and will be promoted to int when passed. Not sure if you want to add a few more details or a reference to a specification here.

craigcomstock (changes requested):
one typo to fix, and wonder about being more detailed in why this change.

libutils/logging.h
larsewi: Added a bit more details. It's quite confusing, but so is variadic arguments in general. It's more of a hack

PR #232, opened 2024-12-18
CMake support kickstart

vpodzime:


larsewi (approved):
Very nice :rocket:

larsewi:
We should maybe have a GitHub Action workflow that does a smoke test by trying to build the library with cmake?

PR #234, opened 2025-01-23
Sparse file copying no longer aborts if fallocate is unsupported

larsewi:
On my Ubuntu 24 VM, I experienced that FileSparseCopy() kept getting
aborted with the following error message:

```
   error: Failed to pre-allocate space for 'promises.cf.cfnew': Operation not supported
```

In the man pages I saw that the system call fallocate(2) can fail if the
file system does not support it. However, fallocate(2) is nothing more
than an optimization. It is not strictly needed. Hence, the file copying
should proceed if fallocate(2) is unsupported.

[Build Status](https://ci.cfengine.com/job/pr-pipeline/11639/)

### DIFF
--- a/libutils/file_lib.c
+++ b/libutils/file_lib.c
@@ -1710,9 +1710,12 @@ bool FileSparseCopy(int sd, const char *src_name,
      */
     if (fallocate(dd, FALLOC_FL_KEEP_SIZE|FALLOC_FL_PUNCH_HOLE, 0, in_sb.st_blocks * 512) != 0)
     {
-        Log(LOG_LEVEL_ERR, "Failed to pre-allocate space for '%s': %m", dst_name);
-        *total_bytes_written = 0;
-        return false;
+        Log((errno == EOPNOTSUPP) ? LOG_LEVEL_VERBOSE : LOG_LEVEL_ERR,
+            "Failed to pre-allocate space for '%s': %m", dst_name);
+        if (errno != EOPNOTSUPP) {
+            *total_bytes_written = 0;
+            return false;
+        }
     }
 
     off_t in_pos = 0;


libutils/file_lib.c
```diff
@@ -1710,7 +1710,16 @@ bool FileSparseCopy(int sd, const char *src_name,
...
     if (fallocate(dd, FALLOC_FL_KEEP_SIZE|FALLOC_FL_PUNCH_HOLE, 0, in_sb.st_blocks * 512) != 0)
     {
-        Log(LOG_LEVEL_ERR, "Failed to pre-allocate space for '%s': %m", dst_name);
+        if (errno == EOPNOTSUPP) {
+            Log(LOG_LEVEL_DEBUG,
+                "Operation to pre-allocate space for '%s' failed with: %m. "
+                "Falling back to less modern system calls", dst_name);
+            return FileSparseCopyShoveling(sd, src_name, dd, dst_name,
+                                           blk_size, total_bytes_written,
+                                           last_write_was_a_hole);
```
vpodzime, line 1719: Hmm, now that I'm thinking about this I wonder if it's not better to just keep going here without the pre-allocation. It's an optimization, basically, we don't strictly needed. The goal is to tell the FS more about what we are about to do and give it a chance to tell us there's not enough space, for example. But if we skip this, it's still as good/bad as `FileSparseCopyShoveling()` is. So we might as well just log a verbose message (warning would be too noisy) and continue with the code below.

vpodzime (commented):
Thanks for fixing this!

libutils/file_lib.c
larsewi: That seems to also work :+1:

craigcomstock (approved):
fair enough. thanks!

PR #236, opened 2025-02-06
CMake JSON

vpodzime:
Adding JSON support as another thing that can be built with CMake plus some small changes and a new GH action.


examples/json_example.c
```diff
@@ -0,0 +1,35 @@
...
+        printf("Error when parsing JSON data: %s\n", JsonParseErrorToString(err));
+        return 1;
+    } else {
+        Writer *stdout_writer = FileWriter(stdout);
+        if (stdout_writer == NULL) {
+            printf("Failed to create new writer: %m\n");
+            JsonDestroy(json);
+            return 1;
+        }
+        JsonWrite(stdout_writer, json, 0 /* base indent level */);
```
larsewi, line 19: Still need to destroy the JSON here :thinking:

examples/json_example.c
```diff
@@ -0,0 +1,35 @@
...
+        return 0;
+    }
+}
+
+int main() {
```
larsewi, line 24:
```suggestion
int main(void) {
```

larsewi (dismissed):
Very nice :rocket:

examples/json_example.c
vpodzime: Good catch! Examples must be correct!

larsewi (approved):

PR #238, opened 2025-03-03
Add static-check and fix a few issues

craigcomstock:
blocks CFE-3429 and https://github.com/cfengine/core/pull/5702

### DIFF
--- a/.github/workflows/static_check.yml
+++ b/.github/workflows/static_check.yml
@@ -1,0 +1,21 @@
+name: Static Check
+
+on:
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  static_check:
+    runs-on: ubuntu-24.04
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        submodules: recursive
+
+    - name: Prepare Environment
+      run: |
+        sudo apt-get update && \
+        sudo apt-get install -y buildah
+
+    - name: Run The Test
+      run: ./tests/static-check/run.sh
--- a/libcompat/inet_pton.c
+++ b/libcompat/inet_pton.c
@@ -83,7 +83,8 @@ inet_pton4(src, dst)
 {
 	static const char digits[] = "0123456789";
 	int saw_digit, octets, ch;
-	unsigned char tmp[NS_INADDRSZ], *tp;
+        unsigned char tmp[NS_INADDRSZ] = {0};
+        unsigned char *tp;
 
 	saw_digit = 0;
 	octets = 0;
--- a/tests/static-check/Makefile.am
+++ b/tests/static-check/Makefile.am
@@ -1,0 +1,25 @@
+#
+#  Copyright 2025 Northern.tech AS
+#
+#  This file is part of CFEngine 3 - written and maintained by Northern.tech AS.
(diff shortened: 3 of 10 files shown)


tests/static-check/Makefile.am
```diff
@@ -0,0 +1,25 @@
+#
+#  Copyright 2021 Northern.tech AS
```
larsewi, line 2:
```suggestion
#  Copyright 2025 Northern.tech AS
```

.github/workflows/static_check.yml
```diff
@@ -0,0 +1,25 @@
...
+    runs-on: ubuntu-24.04
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        submodules: recursive
+
+    - name: Prepare Environment
+      run: |
+        sudo apt-get update && \
+        sudo apt-get install -y dpkg-dev debhelper g++ libncurses6 pkg-config \
```
larsewi, line 18: Do we really need `debhelper` for static checks?

.github/workflows/static_check.yml
```diff
@@ -0,0 +1,25 @@
...
+          build-essential fakeroot gcc make autoconf buildah
+
+    - name: Run Autogen
+      run: ./autogen.sh --enable-debug --with-pcre2=no --with-openssl=no --with-libyaml=no
```
larsewi, line 22: Why build without PCRE2, OpenSSL and libyaml? Maybe worth a comment?

libcompat/inet_pton.c
```diff
@@ -83,7 +83,8 @@ inet_pton4(src, dst)
 {
 	static const char digits[] = "0123456789";
 	int saw_digit, octets, ch;
-	unsigned char tmp[NS_INADDRSZ], *tp;
+        unsigned char tmp[NS_INADDRSZ] = 0;
```
larsewi, line 86: Array must be initialized with a brace-enclosed initializer, I think
```suggestion
        unsigned char tmp[NS_INADDRSZ] = {0};
```

larsewi (commented):
`GetProcdirPid()` is never used?

.github/workflows/static_check.yml
craigcomstock: neither dpkg-dev probably, will adjust/minimize

larsewi (approved):
:rocket:

tests/static-check/run_checks.sh
```diff
@@ -0,0 +1,51 @@
+#!/bin/bash
+
+set -x
+
+n_procs="$(getconf _NPROCESSORS_ONLN)"
+
+function check_with_gcc() {
+  make clean
+  ./autogen.sh --enable-debug CC=gcc
```
olehermanse, line 9: @craigcomstock This one is outdated now, compared to https://github.com/cfengine/core/pull/5706/files

I think they should be the same and we should work towards how we discussed it to be (this test does not rely on autotools - a separate step beforehand should prepare a ready-to-use source tarball where we run configure).

craigcomstock: yeah, will fix.

PR #239, opened 2025-03-06
Moved GetRelocatedProcdirRoot() from libntech to core

craigcomstock:
Commit in core where this was moved is cfengine/core@3d914b512a9e92e0a06068934199f026c3761cd0

Ticket: CFE-3429

### DIFF
--- a/libutils/string_lib.c
+++ b/libutils/string_lib.c
@@ -374,7 +374,7 @@ char *StringConcatenate(size_t count, const char *first, ...)
     char *result = xcalloc(total_length + 1, sizeof(char));
     if (first)
     {
-        strcat(result, first);
+        strncpy(result, first, total_length);
     }
 
     va_start(args, first);
--- a/libutils/file_lib.c
+++ b/libutils/file_lib.c
@@ -1935,21 +1935,6 @@ ssize_t CfReadLines(char **buff, size_t *size, FILE *fp, Seq *lines)
 
 /*******************************************************************/
 
-const char* GetRelocatedProcdirRoot()
-{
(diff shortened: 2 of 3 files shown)


larsewi (dismissed):

larsewi (approved):

libutils/string_lib.c
```diff
@@ -374,7 +374,7 @@ char *StringConcatenate(size_t count, const char *first, ...)
     char *result = xcalloc(total_length + 1, sizeof(char));
     if (first)
     {
-        strcat(result, first);
+        strncat(result, first, total_length);
```
larsewi, line 377: > If the destination buffer does not already contain a string, or is not large enough, the behavior is undefined.

 - Ref. `man strncat`

Ops, it seems that there must be a string already in `result`, which there is not. Maybe `strncpy` is the correct function to use here? I.e., copy the first string into `dst`, then concatenate the rest of them.

larsewi (changes requested):

libutils/string_lib.c
craigcomstock: well darn! on debian-12 `man strncat` doesn't mention the caveat!

```
Linux man-pages 6.03                                        2023-02-05                                                  strncat(3)
```

Checking online I see linux man-pages 6.10/2024-07-23 says what you quote.

On my system with 6.03/2023-02-05 it says

```
       If the destination buffer is not large enough, the behavior is undefined.  See _FORTIFY_SOURCE in feature_test_macros(7).
```

eek! You might think this was a bit more stable than that! ;)

larsewi (approved):
Looks great :rocket:

PR #241, opened 2025-03-11
Fixed static-check in case where autogen has not run, to not configure

craigcomstock:


tests/static-check/run_checks.sh
```diff
@@ -6,7 +6,7 @@ n_procs="$(getconf _NPROCESSORS_ONLN)"
 
 function check_with_gcc() {
   make clean
-  ./autogen.sh --enable-debug CC=gcc
+  ./autogen.sh --enable-debug CC=gcc NO_CONFIGURE=1
```
larsewi, line 9: `NO_CONFIGURE=1` needs to come before `./autogen.sh`. At least on my machine :wink:

larsewi (changes requested):

craigcomstock:
wrong pr, will re-do :)

PR #242, opened 2025-03-11
Fixed static-check to run autogen NO_CONFIGURE if needed and then configure in each test

craigcomstock:
Running autogen.sh in each test is inefficient.

We don't run this test direclty on libntech in jenkins but I included those comments to keep the files more in sync between core and libntech.


vpodzime (approved):
ACK

larsewi (approved):

PR #243, opened 2025-03-26
CFE-4385: Fix build on macOS

btriller:
Check for Homebrew if we're on Darwin and use brew's --prefix. Also use autoconf's macro for egrep to get rid of obsolescent warnings.

Ticket: CFE-4385

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -88,9 +88,22 @@ dnl ######################################################################
 dnl Checks for programs.
 dnl ######################################################################
 
+AC_PROG_GREP
+AC_PROG_EGREP
 AC_PROG_CC
 AC_PROG_MKDIR_P
 AC_EXEEXT
+
+AS_CASE([${target_os}],
+[darwin*], [
+    AC_CHECK_PROG([BREW], [brew], [brew])
+    AC_MSG_CHECKING([for OS X Homebrew])
+    AS_IF([test "x$BREW" = "xbrew"], [
+            have_brew=yes
+            AC_MSG_RESULT([OS X Homebrew detected])
+        ],
+        [have_brew=])
+])
 
 dnl GCC specific flags
 m4_include([m4/cf3_gcc_flags.m4])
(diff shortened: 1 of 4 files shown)


larsewi:
@cf-bottom Jenkins with exotics please :)

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/11917/)

(with exotics)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/11917/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-11917/

configure.ac
```diff
@@ -88,10 +88,23 @@ dnl ######################################################################
...
 AC_PROG_MKDIR_P
 AC_EXEEXT
 
+AS_CASE([${target_os}],
+[darwin*], [
+    AC_CHECK_PROG([BREW], [brew], [brew])
+    AC_MSG_CHECKING([for OS X Homebrew])
+    AS_IF([test "x$BREW" = "xbrew"], [
+            have_brew=yes
+            AC_MSG_RESULT([OS X Homebrew detected])
```
craigcomstock, line 103: Seems like we should have a "Homebrew not detected." I haven't used macports in a while but that would be the other reasonable alternative right?

craigcomstock (commented):
Cool. Maybe add a "no homebrew detected" message and maybe bomb out if darwin? Or maybe check for macports. Not sure if macports would need special LDFLAGS/CPPFLAGS or not...

configure.ac
olehermanse: @craigcomstock In this case, the only reason why we care about brew, is because it can give us the path to various libraries like OpenSSL etc. I haven't heard of too many people using macports, I'd guess the other, most likely "alternative" (IMO), is no package manager, i.e. finding the libraries in some standard system paths, or by manually supplying them as arguments to configure.

craigcomstock: Right on. Last time I worked with another developer who used macOS they used macports. Still an else saying homebrew is required would be good.

olehermanse: @craigcomstock but it isn't required? It's just one more way / location to look for the libraries. You can specify them via command line arguments or have the libraries in whatever is the default location(s) on the system(?).

craigcomstock: Yeah, maybe just a suggestion: "No Homebrew detected. You will need to provide command line arguments LDFLAGS/CPPFLAGS to provide locations for required libraries: rsync, openssl, ..."

PR #244, opened 2025-03-28
Adjusted logic for OS X brew and --with options to configure

olehermanse:
When a user specifies --with-something=/path/to/something in ./configure we don't want that to be overridden. In other words, we only want to do the automatic brew --prefix checking if the with_something variable is yes or check.


craigcomstock (approved):
Outstanding! Good catch.

PR #246, opened 2025-04-01
CFE-3629 - just poking around

btriller:
Set CFLAGS before AC_PROG_CC as it might add -g/-O options if CFLAGS is unset.
Don't add CFLAGS to CORE_FLAGS because I don't know, but maybe CFLAGS is populated otherwise.

### DIFF
--- a/autogen.sh
+++ b/autogen.sh
@@ -53,9 +53,9 @@ test -z "$srcdir" && srcdir=.
 cd "$srcdir"
 
 echo "$0: Running autoreconf ..."
-autoreconf -Wno-portability --force --install -I m4  ||  exit
+autoreconf -Wno-portability --force --install -I m4 || exit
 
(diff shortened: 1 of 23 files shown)


olehermanse:
@btriller Why did you close this? IIRC, the change looked reasonable, but more work was needed to fix the failing tests(?)

btriller:
> @btriller Why did you close this? IIRC, the change looked reasonable, but more work was needed to fix the failing tests(?)

I looked further into it on macOS (with clang). With `-Wall -Werror`, there were many errors about redefinition that I wasn't sure I'm able to resolve it. IIRC on Linux with gcc and, for starters, more ifdefs in libcfecompat, it looked better.

olehermanse:
@btriller Thanks for working on this. We're a bit busy with getting the releases out, but I'll get back to reviewing the PR shortly after that :)

PR #247, opened 2025-04-04
CFE-1484: Added module_dir sys variable

victormlg:


libutils/known_dirs.c
```diff
@@ -134,6 +135,21 @@ const char *GetWorkDir(void)
...
 
+const char *GetModuleDir(void)
+{
+    const char *module_dir = getenv("CFENGINE_TEST_OVERRIDE_MODULE_DIR");
+
+    if (module_dir == NULL) 
+    {
+        const char* workdir = GetWorkDir(); 
+        char *default_module_dir = strdup(workdir);
+        PathAppend(default_module_dir, PATH_MAX, "modules", '/');
```
larsewi, line 146: The `strdup()` function will allocate `strlen()` bytes to the heap-allocated buffer. But you are telling the `PathAppend()` that the buffer is `PATH_MAX` bytes. Hence, it will write outside the bounds of the buffer, causing undefined behavior. Instead make a buffer `PATH_MAX` bytes large and copy the first string into it before calling `PathAppend()`

larsewi (dismissed):

libutils/known_dirs.c
larsewi: Also use FILE_SEPARATOR macro to make it platform independent. On windows backslash is used as path separator

larsewi: Maybe `snprintf()` is an easier function to use here?

olehermanse:
@cf-bottom jenkins, please

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/12007/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/12007/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-12007/

libutils/known_dirs.c
```diff
@@ -25,6 +25,7 @@
 #include <known_dirs.h>
```
olehermanse, line 25: See comments on https://github.com/NorthernTechHQ/libntech/pull/249 - same applies here.

libutils/known_dirs.c
```diff
@@ -25,6 +25,7 @@
...
 #include <definitions.h>
 #include <file_lib.h>
+#include <string_lib.h>
```
olehermanse, line 28: Why is this include necessary?

PR #249, opened 2025-04-09
CFE-2822: Added ppkeys sys variable

victormlg:


craigcomstock (approved):

libutils/known_dirs.c
```diff
@@ -159,6 +159,23 @@ const char *GetBinDir(void)
...
+        const char* workdir = GetWorkDir(); 
+        char dest[PATH_MAX];
+        snprintf(dest, PATH_MAX, "%s" FILE_SEPARATOR_STR "%s", workdir, "ppkeys");
+        (*allocated) = true;
+
+        return strdup(dest);
+    }
+    (*allocated) = false;
+    return keydir;
+}
```
olehermanse, line 177: @victormlg Why is this one so different from all the other ones?

All the other similar functions seem to use `GET_DEFAULT_DIRECTORY_DEFINE` macro and do not have a `allocated` parameter.

craigcomstock: I think if you do it the "as-is way" you may need to adjust configure.ac as we provide configuration options to change the location of the various directories.

craigcomstock (changes requested):
Consider using the "as-is way" of getting dirs and also of providing configuration options to override the defaults.

olehermanse (approved):

PR #255, opened 2025-08-18
ENT-13168: Specified the deployment target to be macOS 15.4

larsewi:


craigcomstock (approved):
So we have to keep changing this hard coded value maybe? When other functions we use are introduced. Doesn't seem horrible but feels kind of weird.

Could we instead query what the latest is installed and hard-code for that?

https://github.com/actions/runner-images/blob/macos-15-arm64/20250811.2170/images/macos/macos-15-arm64-Readme.md#installed-sdks

mentions that 15.5 SDK is available as well as 26.0 (beta).

Maybe this command we could use to force the latest?

```
$ clang --version
Apple clang version 17.0.0 (clang-1700.0.13.5)
Target: x86_64-apple-darwin24.6.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin
```

PR #268, opened 2026-04-27
Fix sparse copy

btriller:
Save unsupported operation error so that correct error is used after Log()'ing.

Some filesystems (e.g. pmxcfs [1]) do not support fallocate(), so standard way for copying files (cf-before-edit/cf-after-edit) around will fail on these filesystems. This is fine as copying might still succeed. But to proceed we need to check returned error code, which might get overwritten by other function calls.

[1] https://pve.proxmox.com/wiki/Proxmox_Cluster_File_System_(pmxcfs)

### DIFF
--- a/libutils/file_lib.c
+++ b/libutils/file_lib.c
@@ -1710,9 +1710,10 @@ bool FileSparseCopy(int sd, const char *src_name,
      */
     if (fallocate(dd, FALLOC_FL_KEEP_SIZE|FALLOC_FL_PUNCH_HOLE, 0, in_sb.st_blocks * 512) != 0)
     {
-        Log((errno == EOPNOTSUPP) ? LOG_LEVEL_VERBOSE : LOG_LEVEL_ERR,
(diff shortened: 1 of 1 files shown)


libutils/file_lib.c
```diff
@@ -1710,9 +1710,10 @@ bool FileSparseCopy(int sd, const char *src_name,
      */
     if (fallocate(dd, FALLOC_FL_KEEP_SIZE|FALLOC_FL_PUNCH_HOLE, 0, in_sb.st_blocks * 512) != 0)
     {
-        Log((errno == EOPNOTSUPP) ? LOG_LEVEL_VERBOSE : LOG_LEVEL_ERR,
+        int opnotsupp = errno == EOPNOTSUPP;
```
larsewi: Could you please change the type to `bool`? Also, maybe a variable name like, `not_supported` or `operation_not_supported` would read a little better? It's up to you 😉

larsewi (commented):
Thanks 🚀

larsewi (approved):

PR #284, opened 2026-07-16
CFE-3882: Implemented date parser

victormlg:
Probably not the most optimal implementation, but works for most usecases. I tried followed the specs from [gnu date input format](https://www.gnu.org/software/coreutils/manual/html_node/Date-input-formats.html) and inspired myself from the behavior of `$ date --date=""` when in doubt.

Some comments:
- unsure if the timezone math is correct, but it looks ok to me (I got a bit confused)
- it only supports UTC and Z as timezones for now
- `@second-until-epoch` is not supported yet

### DIFF
--- a/libutils/date_parser.c
+++ b/libutils/date_parser.c
@@ -1,0 +1,1369 @@
+/*
+  Copyright 2026 Northern.tech AS
+
+  This file is part of CFEngine 3 - written and maintained by Northern.tech AS.
+
+  Licensed under the Apache License, Version 2.0 (the "License");
+  you may not use this file except in compliance with the License.
+  You may obtain a copy of the License at
+
+      http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+
+  To the extent this program is licensed as part of the Enterprise
+  versions of CFEngine, the applicable Commercial Open Source License
+  (COSL) may apply to this file if you as a licensee so wish it. See
+  included file COSL.txt.
+*/
+
+#include <time.h>
+#include <sequence.h>
+#include <stdbool.h>
+#include <string_lib.h>
+#include <ctype.h>
+#include <alloc.h>
+#include <errno.h>
+
+/*
+ * https://www.gnu.org/software/coreutils/manual/html_node/Date-input-formats.html
+ */
+
+typedef enum {
+    DAY_SUNDAY = 0,
+    DAY_MONDAY,
+    DAY_TUESDAY,
+    DAY_WEDNESDAY,
+    DAY_THURSDAY,
+    DAY_FRIDAY,
+    DAY_SATURDAY,
+} Weekday;
+
+
+typedef enum {
+    MONTH_JANUARY = 0,
+    MONTH_FEBRUARY,
+    MONTH_MARCH,
+    MONTH_APRIL,
+    MONTH_MAY,
+    MONTH_JUNE,
+    MONTH_JULY,
+    MONTH_AUGUST,
+    MONTH_SEPTEMBER,
+    MONTH_OCTOBER,
+    MONTH_NOVEMBER,
+    MONTH_DECEMBER
+} Month;
+
+
+typedef enum {
+    MERIDIAN_AM,
+    MERIDIAN_PM
+} Meridian;
+
+
+typedef enum {
+    ORD_THIS = 0,
+    ORD_FIRST = 1,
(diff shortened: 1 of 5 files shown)


libutils/date_parser.c
```diff
@@ -0,0 +1,1368 @@
...
+    MERIDIAN_AM,
+    MERIDIAN_PM
+} Meridian;
+
+
+typedef enum {
+    ORD_THIS = 0,
+    ORD_FIRST = 1,
+    ORD_THIRD = 3,
+    ORD_FOURTH,
```
olehermanse, line 73: Comment to explain why 2 is missing?

olehermanse (dismissed):

libutils/date_parser.c
```diff
@@ -0,0 +1,1369 @@
...
+    }
+    return false;
+}
+
+
+static Seq *Tokenize(const char *data)
+{
+    Seq *tokens = SeqNew(10, free);
+
+    DateToken token = {0};
```
larsewi, line 343: You should probably allocate it on the heap to begin with.

libutils/date_parser.c
```diff
@@ -0,0 +1,1369 @@
...
+static DateToken *NextToken(ParserState *state)
+{
+    if (state->error || state->idx + 1 >= SeqLength(state->tokens))
+    {
+        state->idx = SeqLength(state->tokens);
+        return NULL;
+    }
+    return (DateToken *) SeqAt(state->tokens, ++state->idx);
+
+}
```
larsewi, line 472: Could reuse function above?
```suggestion
static inline DateToken *NextToken(ParserState *state)
{
    return PeekTokenAhead(state, 1);
}
```

libutils/date_parser.c
```diff
@@ -0,0 +1,1369 @@
...
+
+static int64_t ToYear(int64_t year, size_t ndigit)
```
larsewi, line 474: Would be nice to have a comment here to explain what this function is supposed to do.

libutils/date_parser.c
```diff
@@ -0,0 +1,1369 @@
...
+    {
+        year += 2000;
+    }
+    return year - 1900;
+}
+
+/* =========================== */
+
+
+static bool ParseInt(DateToken *token, int64_t *out)
```
larsewi, line 495: `token` should be `const`?
```suggestion
static bool ParseInt(const DateToken *token, int64_t *out)
```

libutils/date_parser.c
```diff
@@ -0,0 +1,1369 @@
...
+{
+    assert(token->type == DATE_TOKEN_INTEGER);
```
larsewi, line 497:
```suggestion
    assert(token != NULL);
    assert(token->type == DATE_TOKEN_INTEGER);
```

libutils/date_parser.c
```diff
@@ -0,0 +1,1369 @@
...
+{
+    if (IsEnd(state))
+    {
+        return NULL;
+    }
+    return (DateToken *) SeqAt(state->tokens, state->idx);
+}
+
+
+static DateToken *PeekTokenAhead(ParserState *state, size_t ahead)
```
larsewi, line 452: I feel like these functions should return `const DateToken *`. You generally don't want to modify the tokens in any way when parsing text

tests/unit/date_parser_test.c
```diff
@@ -0,0 +1,188 @@
...
+    assert(t == 1789948800);
+    assert(StringToTime("September 21, 2009", &t) == 0);
+    assert(t == 1253491200);
+    assert(StringToTime("September 21 2009", &t) == 0);
+    assert(t == 1253491200);
+
+    assert(StringToTime("September 21 09", &t) == 0);
+    assert(t == 1253491200);
+#if !defined(__APPLE__) && !defined(__FreeBSD__) && !defined(__OpenBSD__) \
+    && !defined(__NetBSD__)
```
larsewi, line 40: Please add a comment explaining why this is here

tests/unit/date_parser_test.c
```diff
@@ -0,0 +1,188 @@
...
+    time_t t;
+    assert(StringToTimeFrom("10:10", &t, &mydate) == 0);
+    assert(t == 1783246200);
+    assert(StringToTimeFrom("10:10:10", &t, &mydate) == 0);
+    assert(t == 1783246210);
+    assert(StringToTimeFrom("10:10:10.00001", &t, &mydate) == 0);
+    assert(t == 1783246210);
+    assert(StringToTimeFrom("10:10 am", &t, &mydate) == 0);
+    assert(t == 1783246200);
+    assert(StringToTimeFrom("10:10 p.m.", &t, &mydate) == 0);
```
larsewi, line 73: What happens if?
```suggestion
    assert(StringToTimeFrom("10:10 p.m. am", &t, &mydate) == 0);
```

larsewi (commented):
Did you consider using YACC/Bison grammar?

PR #288, opened 2026-08-05
ENT-8193: Added validfiledata policy function

victormlg:


libutils/json-utils.c
```diff
@@ -320,7 +320,7 @@ bool JsonParseCsvFile(const char *input_path, size_t size_max, JsonElement **jso
 }
 
 JsonElement *JsonReadDataFile(const char *log_identifier, const char *input_path,
-                              const DataFileType requested_mode, size_t size_max)
+                              const DataFileType requested_mode, size_t size_max, bool errors)
```
larsewi: This change does not add validfiledata policy function. The commit message should instead explain what this new argument does. You can mention that it's needed for the validfiledata policy function.

victormlg:
> Maybe it would be better to create another function that is called `IsDataFileValid`? Did you tests that `JsonParseEnvFile` and `JsonParseCsvFile` don't print errors?
> 
> It's strange if this function only prints errors for JSON.

I moved the error logging from JsonParseEnvfile and JsonParseCsvFile to the outer function JsonReadDatafile, so that when calling validfiledata() in the policy language, we ignore all error message.

libutils/json-utils.c
```diff
@@ -320,7 +314,7 @@ bool JsonParseCsvFile(const char *input_path, size_t size_max, JsonElement **jso
...
```
larsewi:
```suggestion
                              const DataFileType requested_mode, size_t size_max, bool ignore_errors)
```

libutils/json-utils.c
```diff
@@ -341,6 +335,15 @@ JsonElement *JsonReadDataFile(const char *log_identifier, const char *input_path
         }
         if (success == false)
         {
+            if (!ignore_errors)
+            {
+                return NULL;
+            }
+            const char *file_type = (env_mode) ? "ENV" : "csv";
+            const char *myname = (env_mode) ? "JsonParseEnvFile" : "JsonParseCsvFile";
```
larsewi:
```suggestion
            const char *file_type = env_mode ? "ENV" : "csv";
            const char *myname = env_mode ? "JsonParseEnvFile" : "JsonParseCsvFile";
```

